PR #113, opened 2015-10-19
Publisher: much faster check if installed on AliEn

dberzano:
Using a single "alien find" rather than an "alien ls" per package.

Thanks Miguel Martinez Pedreira!


mimarpe:
Nice :-)
For the record, I do a simple "find /alice/packages *" to register them.

Cheers,
    Miguel

PR #227, opened 2016-04-22
Publish SHERPA & POWHEG

ktf:


ktf:
@qgp this hopefully takes care of it.

PR #235, opened 2016-04-26
Make .*/root6 branch derived from .*/next

ktf:
As discussed in alisw/alidist#337, we need to derive the root6 branch from next, so that it can be useful also for O2 developments.


ktf:
@barthelemy: I will merge this tomorrow after discussing it with @pzhristov and @dberzano .

PR #251, opened 2016-05-13
Align IB/v5-08/o2 to IB/v5-08/root6

ktf:


ktf:
See alisw/alidist#337.

PR #266, opened 2016-06-21
Make sure mesos works with packer generated docker images

ktf:


ktf:
@kvnnap this should fix your issue.

PR #274, opened 2016-06-23
Improve legacy AliEn handling in CVMFS

dberzano:
* Added function that loads AliEn if not loaded as a dependency for a restricted number of packages
* Load AliEn after other packages to prevent conflicts with GCC-Toolchain


dberzano:
@qgp: bizarrely enough, this should fix the problem you were experiencing on lxplus. Will push it to CVMFS soon.

PR #282, opened 2016-06-24
Fix Jenkins test on branch

dberzano:


dberzano:
Jenkins test _correctly_ fails as the version of `alienv` currently in the repo is effectively broken in the `printenv` function.

PR #284, opened 2016-07-06
Fix alienv printenv issue

dberzano:


dberzano:
This is the version currently deployed on CVMFS. As expected, it passes the Jenkins test. Merging.

PR #288, opened 2016-07-12
Add check to AliEn PackMan output

dberzano:
AliEn PackMan might silently return zero packages from time to time. This is now considered an error condition and does not trigger any longer the republishing of all packages.


dberzano:
@costing this should fix the problem seen on `alimonitor`.

PR #291, opened 2016-07-20
Add Titan support to alienv

dberzano:


dberzano:
@pavlo-svirin

dberzano:
Tomorrow morning Geneva time (when we can react in case something's wrong) we will port this file to the Grid. All the stress tests passed. Until then we won't merge this PR.

PR #298, opened 2016-07-29
Allow devel packages for AliRoot and AliPhysics

ktf:


ktf:
Required to get the automated coverage reports.

PR #335, opened 2016-10-25
Add Ubuntu 16.* support to alienv

dberzano:


dberzano:
@ktf, @mimarpe, @costing: that this allows `alienv` to list and load correctly the available packages, however it does not necessarily imply that they can be used. For them to work we should provide our GCC compiled for Ubuntu 16.04+. In any case this is enough to make the `alimonitor` problem disappear.

PR #341, opened 2016-11-02
Publish GCC-Toolchain for Ubuntu 16.04

dberzano:
Also prevents to pollute CVMFS with plenty of useless GCC versions by specifying
a fixed one instead of a broad regexp.


dberzano:
@pzhristov once merged this would allow you to run `gdb` from Ubuntu 16.04 (and should also solve some Grid-wide problems for those using modern OSes on nodes).

PR #359, opened 2016-11-29
Improve tolerance on CVMFS delays handling

dberzano:
We now consider the last time the Stratum-0 was updated instead of blindly
computing the difference between time revisions in order to avoid plenty of
false positives.

PR #360, opened 2016-11-30
Improve CVMFS monitor notify logic

dberzano:
Current logic matches the one from the webpage[1]. Service is considered:

* OK if Stratum-1 and Stratum-0 have the same snapshot
* Syncing if Stratum-1 is behind less than `max_revdelta` revisions and
  Stratum-0 changed less than `max_timedelta` seconds ago
* Error if Stratum-1 is outdated, either too many revisions or more than
  `max_timedelta` seconds

Logic introduced by #359 is fine but way too tolerant: in the case of the ALICE
OCDB for instance updates are very frequent so notifications were never
triggered even if the Stratum-1 was too many revisions behind.

[1] http://cernvm-monitor.cern.ch/cvmfs-monitor/alice-ocdb.cern.ch/

### DIFF
--- a/cvmfs-mon/cvmfs-mon.py
+++ b/cvmfs-mon/cvmfs-mon.py
@@ -17,6 +17,14 @@ def notify(notif, to, **keys):
   subject = notif["subject"] % keys
   body = "Subject: %s\nFrom: %s\nTo: %s\n\n" % (subject, notif["from"], ", ".join(to))
   body += notif["body"] % keys
+  if os.environ.get("CVMFSMON_NO_NOTIF", "0") != "0":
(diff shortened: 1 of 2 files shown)


dberzano:
@mimarpe, @maarten-litmaath: this iteration is better (same logic as the [webpage](http://cernvm-monitor.cern.ch/cvmfs-monitor/alice-ocdb.cern.ch/)).

* Notifications are sent every 4 hours
* Repositories are checked every 30 minutes
* A repository is outdated if more than 3600 s behind the Stratum-0 (if revisions differ) or more than 4 revisions behind (which should catch OCDB errors)

We should bear in mind that OCDB updates are supposed to be much more frequent than packages.

PR #363, opened 2016-12-13
Ali titan

pavlo-svirin:
AliRoot build request for ORNL Titan


dberzano:
Hello @pavlo-svirin, this pr has several problems:
* it needs to be rebased (do a `git pull --rebase` before pushing)
* it uploads AliRoot only (we need AliPhysics too for running on Titan, and the corresponding AliRoot is taken automatically)
* it does not add the test (see `test-titan.yaml`)
* it adds a regexp that pulls on Titan an sh*tload of versions of AliRoot (and not only the latest as discussed privately)

I am opening an amended pull request where I pull the _latest_ AliRoot only. I'll tag you as a reference for next publishing requests :-)

PR #366, opened 2016-12-21
Add RPMs for Configuration and O2 for DAQ

dberzano:


dberzano:
@Barthelemy let's wait for the tests. When done, then we'll have O2 RPMs for the default `o2-daq`.

update-run.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -x
```
ktf: Shouldn't this run with `-e`?

PR #381, opened 2017-01-24
Add caching to aliPublish

dberzano:
JSONs coming from URLs containing dependencies that never change are cached on
disk if the optional `--cache-deps-dir` parameter is supplied.


dberzano:
@pavlo-svirin I have improved quite dramatically the time required by aliPublish to run by implementing local caching for HTTP requests that are supposed to always return the same response. Can you test whether this iteration still works on Titan please?

pavlo-svirin:
Tested on Titan, works perfectly, thanks

PR #389, opened 2017-02-14
Software list on Titan

pavlo-svirin:
Please add the pull request with changes to software list on Titan


dberzano:
Hi @pavlo-svirin, from what I understand you need to publish v5-08-21 on Titan, right? For that, you need to publish the toplevel AliPhysics package v5-08-21-01. For the future, you need to know that all AliPhysics versions have a "-01" appended (weird convention I agree, not my call!) and they are the toplevel package, bringing all the dependencies (including AliRoot and GEANT4) with them.

Before making a pull request you should maybe sync your current local branch on your laptop with the master. For this:

```
cd ali-bot/
git fetch --all
git reset --hard origin/master
git clean -fxd
```

Then you can do your modifications to `publish/aliPublish-titan.conf` in the AliPhysics section. Those are regular expressions.

Before doing the pull request, you can also have a look at `publish/test-titan.yaml`. It contains a list of packages with `True` if they are supposed to be published (_i.e._ match the regular expression from the configuration), `False` if they are not. When you open a pull request, an automatic test will check if the configuration gives the results expected from the test.

Anyway when you're done with your edits you can push your changes upstream:

```
git push -f
```

where the `-f` (for "force") is necessary this time. I am assuming you are pushing changes to _your_ fork, from which you can open the pull request.

Anyway have a look at #390 and confirm that one works for you!

PR #407, opened 2017-02-27
Timeout for N second or for an HTTP GET request

ktf:
This will sleep N seconds according to the `-t N` option unless
an HTTP GET request is done, in which case it will terminate early.


ktf:
@dberzano can you use this as a "sleep" for the permission checker?

PR #411, opened 2017-03-02
Publish GitHub AliPhysics dailies

dberzano:


dberzano:
@jgrosseo fyi: this means we start publishing `_GITHUB` daily tags from yesterday's (which was manual - today will be automatic and we'll do the stress test first).

PR #415, opened 2017-03-08
Allow installation of aurora on CVMFS

alibuild:


dberzano:
Hi @ktf (you are commenting as @alibuild for some reason!), you can do:

```
aurora: True
```

if you intend to publish every single Aurora build. However, I would restrict it to a specific version pattern to prevent every test build to be published (_e.g._ a limit on `v0.1.2`).

PR #430, opened 2017-03-27
Improve CI responsiveness to new pull requests

brinick:
Make list-branch-pr return any untested PRs
immediately, or if none are available, wait for some max time
before returning old PRs to be re-tested. This statistically should
schedule untested PRs more aggressively.


dberzano:
@brinick it seems there's no change to show, I believe your changes were merged already, is this the case? Can you close?

brinick:
yes, it's already in

PR #439, opened 2017-04-13
Disable analytics in CI

dberzano:


ktf:
Are you sure it's aliBuild analytics which are causing the defunct curls? They should be already off, because the builder runs without a TTY. I suspect it's the analytics for the continuous integration itself.

ktf:
I would check if `report-analytics` leaves behind a curl by accident (does it happen only on @costing 's machines?)

dberzano:
Yes:

```
curl --max-time 5 --user-agent "report-analytics/Unknown version. (Linux; x86-64 slc7) Python/2.7.5" -d cid=alientest02-4-aliphysics_github_ci -d v=1 -d aip=1 -d an=continuous-builder.sh -d t=screenview -d av=ALIBOT_VERSION -d tid=UA-77346950-2 -d cd=looping --silent --output /dev/null https://www.google-analytics.com/collect
```

ktf:
I know that, I meant why it does. It's a copy of the aliBuild one, which does not seem to leak. Maybe the copy does something slightly different.

dberzano:
The problem is subtle. The `report-analytics` executable spawns `curl` and exits immediately, causing `curl` to be adopted by `report-analytics`'s parent, that is - the Thermos executor, the container's PID 1. However, the Thermos executor does not reap zombies, therefore the problem.

ktf:
Ok, but then how about we simply make `report-analytics` synchronous by default? In aliBuild it's asynchronous simply because brew does it that way to minimise impact on the user. In the continuous builder I guess we can actually have it synchronous since only the bot will invoke it.

dberzano:
Yes I fully support it (with a short `curl` timeout as well).

ktf:
+1

ktf:
(I am officially back after Easter)

PR #448, opened 2017-05-24
Add shuttle aliBuild steering script

dberzano:
* For both the test shuttle and the production one
* Uses aliBuild
* When building master it falls back to the latest working version in case it
  does not compile


dberzano:
@chiarazampolli as agreed, this is the version currently installed both on the test shuttle and the production one.

PR #450, opened 2017-05-24
Make git fetch compatible with old git versions

dberzano:


dberzano:
@chiarazampolli fixes the problem with older Git versions on the Shuttle nodes.

PR #454, opened 2017-06-14
Add storage fallback on AliEn publisher

dberzano:
* Get the first best storage elements (not just one)
* Try them in random order until a good one is found

This fix works around storage failures not detected by the MonaLisa health
check.


dberzano:
@ktf @costing we'll be trying it on the AliEn publishing machine and if it works this afternoon we can merge it

PR #456, opened 2017-06-14
CI: PR check reports as unhealthy if stuck

dberzano:


dberzano:
@ktf this should correctly report as not healthy in case the internal process loop is stuck

PR #459, opened 2017-06-19
Add flpproto packages to CVMFS nightlies

dberzano:


dberzano:
Do not merge this yet.

@Barthelemy @awegrzyn @ktf @sawenzel this is all we need to publish your packages on CVMFS, in case you need them.

Barthelemy:
I would say : go for it ! 
if it is already there, it will be easier to try it out when we want/have time.

PR #460, opened 2017-06-21
CI: expose PR base hash to aliBuild

dberzano:


dberzano:
This enables the functionality introduced by @sawenzel in alisw/alidlst#769. Testing it now.

dberzano:
I have tested it and verified by looking at the logs that it exposes the right hash, and does the right thing. I think we can merge it. @ktf @sawenzel any objections?

PR #464, opened 2017-06-28
CI: pass head commit before merge to aliBuild

dberzano:


dberzano:
@sawenzel see comment on alisw/alidist#780.

dberzano:
@ktf you have a point: this is because we kept getting confused between what's base, and what's head.

ktf:
This is github naming. It's quite well explained at:

https://help.github.com/articles/creating-a-pull-request/

where it says:

> When thinking about branches, remember that the base branch is where changes should be applied, the head branch contains what you would like to be applied.

Given this kind of naming is used everywhere on github, I would be consistent.

dberzano:
OK I agree, so let's call "base" the "upstream reference version, and "head" the PR one, and let's all be happy about it :-)

dberzano:
Merging as the new changes comply to @ktf's comment (base, head).

PR #474, opened 2017-08-17
Use pip to get aliBuild and dependencies installed

dberzano:


dberzano:
This should have been done a long time ago. We are finally getting rid of all the dependency problems thanks to `pip`.

PR #475, opened 2017-08-17
Fix pip call in daily script

dberzano:


dberzano:
Of course #474 had a (small) problem.

PR #477, opened 2017-08-31
Improve continuous builder Git and macOS support

dberzano:
* Support `du` and `gdu` (on macOS with Homebrew)
* Fix repos update when pull caused an unwanted merge
* Fetch origin before resetting to an origin branch

### DIFF
--- a/ci/continuous-builder.sh
+++ b/ci/continuous-builder.sh
@@ -18,6 +18,13 @@ export ALIBOT_ANALYTICS_APP_NAME="continuous-builder.sh"
 MIRROR=${MIRROR:-/build/mirror}
 PACKAGE=${PACKAGE:-AliPhysics}
 
+# Pick right version of du
+TMPDU=$(mktemp -d)
+DU=du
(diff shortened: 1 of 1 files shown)


ci/continuous-builder.sh
```diff
@@ -114,6 +127,7 @@ while true; do
     git credential-store --file ~/.git-creds store
     git config --global credential.helper "store --file ~/.git-creds"
 
+    set-github-status -c ${STATUS_REF} -s $CHECK_NAME/pending || true
```
ktf: Are you sure about this? We do not want to reset the status in case a PR is already green, no?

dberzano: If I am not mistaken, `list-branch-pr` excludes PRs with a green state? I did not check it. In any case I have accidentally added this line to the PR, let me remove it for now.

dberzano: I have removed the line in the new iteration.

dberzano:
@brinick I have implemented your suggestion on `du/gdu`. This is the version currently in use on the macOS builder.

PR #480, opened 2017-09-15
Add Slack notificator for AliPhysics dailies

dberzano:
* Start checking after 4pm
* Check every 5 minutes: notify immediately if tag is OK (i.e. both on CVMFS and
  AliEn)
* After 5.30pm send a message every 15 minutes if tag is not OK


dberzano:
This has been in production for more than a week on the Slack channel #publisher.

PR #483, opened 2017-10-06
Pick a failed/reviewed PR to re-test randomly

dberzano:
Current C.I. mechanism favors untested PRs, and in case no PR is untested, it
retests the first one which failed (or the first one which was reviewed), not
giving any chance to other failed PRs to get retested. This PR mitigates the
problem by picking a random failed (or reviewed) PR instead of the first one.

### DIFF
--- a/list-branch-pr
+++ b/list-branch-pr
@@ -315,7 +315,8 @@ if __name__ == "__main__":
                 err, out = getstatusoutput("sleep {0}".format(args.poll_time))
                 if err or timeSince(start) >= args.maxWait:
                     # return whatever we have (may be empty)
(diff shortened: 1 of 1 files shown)


ktf:
@brinick I assume your new builder is immune to this problem, right?

brinick:
Actually, not really, as the heapq has a fixed order.
If an untested i.e. higher priority PR turns up,  the builder stops testing
the current, previously failed PR, and puts it back on the priority queue. 
Which will mean it will end up first again. Thus when the untested PR is done, 
the builder will pop again take the same PR. It is not impossible then to imagine
getting stuck in a loop of untested PR - first failed PR - another untested PR -first....

PR #484, opened 2017-10-09
Add generation of "updatable" RPMs

dberzano:
Along with normal aliPublish RPMs, allowing for multiple versions of the same
package to be installed at the same time, this feature enables the creation
of more standard RPMs, without the version name hardcoded in the package string.

Such RPMs are installable without specifying the version:

    yum install alisw-dim

and updatable the same way.

All updatable RPMs get unpacked under the same prefix (as opposed to a different
prefix for each package). Modulefiles through the `aliswmod` command work
accordingly.

### DIFF
--- a/publish/aliPublish
+++ b/publish/aliPublish
@@ -373,8 +373,9 @@ class AliEnPackMan(object):
 
 class RPM(object):
 
-  def __init__(self, repoDir, publishScriptTpl, connParams, dryRun=False):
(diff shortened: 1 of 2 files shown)


dberzano:
@Barthelemy almost there. We just need to run the final checks before merging. Please do not merge this, it's still a WIP!

dberzano:
...and as a matter of fact it does NOT work for the former use case. I'll fix it tomorrow.

dberzano:
This version seems to work! We just need to discuss separately what packages need to be generated with this new method.

PR #489, opened 2017-11-08
Add Configuration-Benchmark to RPMs

dberzano:
@PascalBoeschoten FYI


dberzano:
Once merged, you'll have to manually do a build with Jenkins (and RPM will also be created automatically). @Barthelemy and @awegrzyn are already using that.

PR #491, opened 2017-11-17
Use alidist defaults for Jenkins daily builds

dberzano:
This allows to move entirely the "latest AliRoot version" configuration to alidist, via proper defaults. Jenkins configuration will not need any update anymore.


dberzano:
Proper Jenkins job is already in place and tested with a dummy build.

PR #511, opened 2017-12-04
Add lxplus7 suggestion to publisher notifications

dberzano:


dberzano:
@Barthelemy: this should clarify how the O2 builds should be used.

PR #513, opened 2017-12-08
Publish EOS data for prod tags as well

dberzano:


dberzano:
This finally snapshots EOS data on CVMFS for all AliPhysics tags, not only daily tags. As agreed with the Grid experts:

* Data on EOS with a `.no_access` extension will be copied to CVMFS and the extension is stripped there. This is to prevent existing broken jobs running with an unpatched AliRoot to successfully fall back on EOS, and this is _in addition_ to the removal of the EOS fallback.
* Snapshots for tags on CVMFS now use symlinks to the proper daily snapshot. Snapshots for tags are created only for versions >= v5-09.

PR #517, opened 2017-12-30
Use POSIX check as option to CVMFS checker

dberzano:
Note: done to work around broken CVMFS monitor webservice at cernvm-monitor.cern.ch during Christmas break. Do not merge yet.

### DIFF
--- a/check-daily-slack.py
+++ b/check-daily-slack.py
@@ -1,4 +1,27 @@
 #!/usr/bin/env python
+
+# check-daily-slack.py -- Check for the daily AliPhysics tag, notify on Slack.
+#
+# This script continuously checks for the presence of the AliPhysics daily tag
+# every 300 seconds, on AliEn and CVMFS. If it's past 4pm it checks for today's
+# daily, if it's before, it checks for yesterday's. In case the tag is found on
+# both AliEn and CVMFS, a successful notification is sent and no further
+# notification will occur (until the next daily tag). In case it's not found,
+# a nagging notification is sent every 15 minutes.
+#
+# Time checks are all explicitly performed using the Geneva time zone, and
+# notifications are sent on a certain Slack channel.
+#
+# Configuration is done via environment variables:
+#
+# SLACK_PRIVATE_URL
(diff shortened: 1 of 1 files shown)


check-daily-slack.py
```diff
@@ -40,21 +40,29 @@ def debug(msg):
...
+    for d in getout(parrot + ["ls", "-1", "/cvmfs/alice.cern.ch"])[0].split("\n"):
+      if getout(parrot + ["bash", "-c", "ls -1 /cvmfs/alice.cern.ch/%s/Modules/modulefiles/%s/%s*" % (d, pkg, ver)])[1] == 0:
+        return True
+    # Fallback on Parrot next time
+    parrot = [ "env",
+               "HTTP_PROXY=DIRECT;",
+               "PARROT_ALLOW_SWITCHING_CVMFS_REPOSITORIES=yes",
+               "PARROT_CVMFS_REPO=<default-repositories>",
+               "parrot_run" ]
   return False
```
brinick: Can this not be simplified?

```python
def is_on_cvmfs_posix(pkg, tag):
   for parrot in ([], ["env", .....]):
      if getout(parrot + ["bash", "-c", "ls -1 /cvmfs/alice.cern.ch/*/Modules/....."])[1] == 0:
         return True
   return False
```

dberzano: Well, the `for` line would become incredibly long. This is why I did not do it that way.

brinick: Well, you could always put them into a variable, broken over several lines,
as in your version and then for-loop over the variable.
But anyhow, this is not very important so let me not hold you up! 😉 
It was just that when I looked at the code I was confused for some seconds
wondering what the range loop was all about.

dberzano: Your solution is more Pythonic (`([], long_variable)`), so I am going for it. I did not like the `range()` either to be honest, this was a dirty Christmas fix!

dberzano:
OK, usage description added and more Pythonic syntax used.

dberzano:
Can we merge this? I think I have addressed the comments.

PR #522, opened 2018-01-16
Add coverage support for the continuous builder

ktf:


ci/continuous-builder.sh
```diff
@@ -138,14 +138,17 @@ while true; do
     # we build.
     mkdir -p sw/BUILD
     find sw/BUILD/ -maxdepth 1 -name "*latest*" -delete
+    # Delete coverage files from one run to the other to avoid
```
dberzano: "from one run to the _next_"

ci/continuous-builder.sh
```diff
@@ -164,6 +167,26 @@ while true; do
...
+
+    # Push the coverage information to codecov
+    COVERAGE_INFO_DIR=$(find sw/BUILD/ -maxdepth 4 -name coverage.info | head -1 | xargs dirname)
+    if [[ ${COVERAGE_INFO_DIR} ]]; then
+      pushd ${COVERAGE_INFO_DIR}
+        COVERAGE_COMMIT_HASH=${ALIBUILD_HEAD_HASH}
+        if [[ $COVERAGE_COMMIT_HASH == 0 ]]; then
+          COVERAGE_COMMIT_HASH=${ALIBUILD_BASE_HASH}
+        fi
+        $TIMEOUT_CMD bash <(curl -s https://codecov.io/bash) -y .codecov.yml                  \
```
dberzano: trim spaces between last char and `\`

dberzano: Use `curl --max-time` to set a reasonable timeout too (`$TIMEOUT_CMD bash` is evaluated at a later stage, so we need to make `curl` timeout first)

ci/continuous-builder.sh
```diff
@@ -164,6 +167,26 @@ while true; do
...
+
+    # Push the coverage information to codecov
+    COVERAGE_INFO_DIR=$(find sw/BUILD/ -maxdepth 4 -name coverage.info | head -1 | xargs dirname)
+    if [[ ${COVERAGE_INFO_DIR} ]]; then
+      pushd ${COVERAGE_INFO_DIR}
+        COVERAGE_COMMIT_HASH=${ALIBUILD_HEAD_HASH}
+        if [[ $COVERAGE_COMMIT_HASH == 0 ]]; then
+          COVERAGE_COMMIT_HASH=${ALIBUILD_BASE_HASH}
+        fi
+        $TIMEOUT_CMD bash <(curl --max-time 600 -s https://codecov.io/bash) -y .codecov.yml \
```
dberzano: We have to handle the case where bash receives garbage data (because of a problem with https://codecov.io/bash): `|| true` at order zero.

ktf: Agreed. Fixed.

PR #523, opened 2018-01-16
Protect against coverage.info file not being there

ktf:


ktf:
Currently broken. Merging.

PR #527, opened 2018-01-19
Add support for CI configuration profiles on macOS

dberzano:


ci/run-continuous-builder.sh
```diff
@@ -49,31 +66,53 @@ cd "$CI_WORK_DIR"
 [[ -d alibuild/.git ]]            || git clone https://github.com/alisw/alibuild
 [[ -d "$PR_REPO_CHECKOUT/.git" ]] || git clone "https://github.com/$PR_REPO" "$PR_REPO_CHECKOUT"
 
-if [[ $1 != --test ]]; then
+# Save Git credentials
+printf "protocol=https\nhost=gitlab.cern.ch\nusername=$GITLAB_USER\npassword=$GITLAB_PASS\n" | \
```
ktf: Doesn't this allow a PR to get our credentials?

dberzano: It does indeed, we have to look for an alternative. At the moment, even if we used a different credential store that does not expose  credentials to a file, any rogue `git` command to that specified server will be executed as our user.

Since aliBuild is the only tool allowed to use such credentials (and not the recipes), we should find a more general way to expose them to aliBuild only, which in turn will strip them when running the very recipes.

PR #535, opened 2018-01-25
Improve Jenkins builds

dberzano:
* GitHub slugs in build-any-ib
* `--fetch-repos` where appropriate


dberzano:
We were missing `--fetch-repos` from the daily builds and build-any-ib (not critical). DA RPMs instead needed it (as they fetch AliRoot master), so in that case it was critical.

PR #544, opened 2018-03-12
Improve stability of continuous integration

dberzano:
* Clean build directories to avoid space usage explosion and inconsistent build results due to previous artifacts
* Clean dead Git references growing indefinitetly
* Add ability to force-test only given commits for debug

### DIFF
--- a/ci/continuous-builder.sh
+++ b/ci/continuous-builder.sh
@@ -76,6 +76,9 @@ report_state started
 while true; do
   report_state looping
 
+  # Run preliminary cleanup command
+  alibuild/aliBuild clean ${DEBUG:+--debug}
+
   # Update all Git repositories (except ali-bot)
   for d in $(find . -maxdepth 2 -name .git -exec dirname {} \; | grep -v ali-bot); do
     pushd $d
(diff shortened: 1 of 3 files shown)


dberzano:
Currently testing on macpromat. Will merge tomorrow in case no more problems pop up.

ktf:
I am still worried about the fact we need aliBuild clean, rather than CMake doing it's job and removing stale targets. Why is that?

dberzano:
@ktf CMake issues depend on each project, and have to be fixed independently on the projects themselves. I am afraid that CMake is however unable to natively, for instance, get rid of target directories that were deleted, making them available to the incremental builds.

aliBuild does reuse the build area on purpose (for allowing incremental builds). The main reason for running `aliBuild clean` is not (only) for working around "faulty" (or "intentionally misbehaving"!) CMake projects, but because we don't want build directories for merged PRs to hang around, making the used size grow indefinitely.

aliBuild correctly leaves them in place by default, but - once built and merged - they can go away. The reason why the approach is more aggressive now is that we _also_ work around "faulty" CMake projects.

Did I answer your question? :-)

dberzano:
For the record: average build time for O2 on macOS went down from 21 minutes to 13 on average, thanks to the branches cleanup.

PR #546, opened 2018-03-14
[WIP] Phase out AliEn publication

dberzano:


dberzano:
@mimarpe can you confirm the following two points?

1. We don't need to register packages in AliEn for real as tarballs, as in all cases they will be used from CVMFS.
2. We don't need to register all packages, but only the toplevel ones, such as AliRoot and AliPhysics (as dependencies are loaded consistently by CVMFS's `alienv`).

This should fix the "dependencies problem" we have seen in production because our toplevel packages never clash and always have different names.

By not telling AliEn the dependencies any longer (CVMFS knows them, beware) we are also bypassing the check that triggered warning emails.

mimarpe:
Ciao Dario,

1. They don't need to be tarballs but AliEn has to be able to _find_ them, so some file/entry has to be under the folder with the package name. That causes though an incoherence, because AliEn will think it can provide the package, while the file is not the real package itself. From the pragmatic point of view, it would work, since indeed we use only CVMFS on sites. And job insertions will work since the package is registered.
2. Again, assuming we use CVMFS, this is correct. If somebody would be to use the classic PackMan then dependencies, and packages, have all to be there, AFAICS...

We could discuss if we drop the support for any other PackMan than CVMFS. Or at least for jAliEn, where in any case only CVMFS implementation is ready. But has to be agreed :-)

You could make a test adding a folder (package) and file (platform) combination for the packages and see if they can be listed with: `packman list -all -force`

dberzano:
Hi, as for the first point: I believe what you mean is that the package has to be seen as a file under the AliEn FC:

    /alice/packages/DPMJET/v3.0.5-alice4-1/Linux-x86_64

where the file name is the "AliEn platform". This is, effectively, a tarball (`.tar.gz`), and what I propose as a first step for phasing out Packman is to start registering "dummy" tarballs with barely nothing inside. We might as well create empty files, the result would be the same.

The effect is to speed up AliEn publication, which is now quite time-consuming (but also, it's not a critical problem): we now need to unpack the aliBuild tarball and repack it using the required Packman directory structure.

The problem - which is urgent to be solved and this is the only solution I can think of - is that:

* aliBuild uses different build numbers that are isolated per architecture,
* Packman has no idea of what a build number is: it sees a package name and a version, and once this is selected, it sees many platforms inside the package

So basically in aliBuild the hierarchy is `platform/package` whereas in Packman it's `package/platform`. The two are utterly incompatible, and the only way to make them work _as of today_ is to publish toplevel packages and pretend packages have no dependencies, because our toplevel packages always have different names (even across architectures).

ktf:
Closing this as it does not merge anymore and it's not clear what the status is.

PR #547, opened 2018-03-14
Work around AliEn Packman problem

dberzano:
This fixes the issues with the notification emails for invalid packages

### DIFF
--- a/publish/aliPublish
+++ b/publish/aliPublish
@@ -740,7 +740,16 @@ def main():
                       help="Write PID to this file and do not run if already running")
   parser.add_argument("--cache-deps-dir", dest="cacheDepsDir", default=None,
                       help="Directory where to cache package dependencies (optional)")
+  parser.add_argument("--override", dest="override", nargs="+",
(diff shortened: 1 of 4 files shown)


dberzano:
@mimarpe so as agreed - for the moment we're just reverting the AliEn publication of ROOT6 packages. All SLC6 packages are *not* published on AliEn Packman at the moment (and the existing ones were cleaned up), so there is no chance for clashes. The better solution to avoid them, as discussed, will be implemented when we have consensus.

mimarpe:
Ok. For the record, after discussed:

1 - Implement alienv checkenv <packages> command to verify correctness of the package combination and avoid bogus executions/results
1.1 - This should be added in principle to alienv enter, or alienv printenv
2 - Package sorting by priority of the packages lists passed to alienv
3 - Add checkenv to the JobAgent before execution of the payload
4 - TBD:
4.1 - only display base packages in alimonitor.cern.ch/packages
4.2 - use checkenv on the packages of waiting jobs (using JOBAGENT entries of the TQ)

PR #548, opened 2018-03-14
Include deps on AliEn Packman too

dberzano:


dberzano:
Forgot to add this in #547.

PR #549, opened 2018-03-14
Improve alienv for AliEn integration

dberzano:
As discussed with @mimarpe.

- [x] Add `checkenv` returning nonzero if deps are conflicting
- [x] Order packages by prioritizing O2, AliPhysics, AliRoot, ROOT and AliGenerators
- [x] Add unittest for `alienv` in Travis CI (start from Jenkinsfile)
- [x] Add ability to switch to "nightlies" (we've decided not to)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 ---
 sudo: required
+cache:
+  directories:
+    - $HOME/cached
 script:
   - |
     git diff --name-only $TRAVIS_COMMIT_RANGE | grep -q ^riemann || return 0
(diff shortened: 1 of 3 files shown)


dberzano:
Hi @mimarpe, I just wanted to decide what's the best way to address the "nightlies" point.

You currently run something like this in the JA:

    /cvmfs/alice.cern.ch/bin/alienv printenv [packages]

I am thinking, as discussed, what is the best way to give ability to pick the package stack from alice-nightlies.cern.ch. We have several options:

1. **The cleanest.** AliEn will have a `$CVMFS_PREFIX` variable that the JA uses, and this will lead to replacing - in AliEn - all hardcoded /cvmfs/alice.cern.ch to `$CVMFS_PREFIX`. Haven't we done something similar for Titan? This would solve all our problems even in places with no CVMFS.

2. **Command-line option.** We add an `alienv -n` option to all our deployed alienvs. Even if we call /cvmfs/alice.cern.ch/bin/alienv, this will exec the other alienv in case. Still, it needs a variable to be exported to steer the correct switch in the JA.

3. **Environment variable.** Same as above, but by exporting a variable.

Honestly: I'd use 1...

mimarpe:
Ciao @dberzano,
most options are basically the same but I'd do 1. too.
Let's use the JDL tag CVMFS_prefix to let the JobAgent override the source of the environment for the packages.
"CVMFS_prefix" = "/cvmfs/alice-nightlies.cern.ch";

Then we will assume all CVMFS repos have alienv in $PREFIX/bin/alienv.

OK? :-)

dberzano:
> `$PREFIX/bin/alienv`

This is a safe assumption, that's what we do!

dberzano:
@awegrzyn @Barthelemy you asked how to add distinct Travis tests according to the files modified by a Pull Request, in this PR you have an example (check the `.travis.yml`).

dberzano:
Ready to be pushed to production.

PR #572, opened 2018-06-07
CI: Generate badges upon default branch state

dberzano:


dberzano:
Tested since months. Merging as discussed during WP3.

PR #573, opened 2018-06-07
Do not use comments in PRs (optional)

dberzano:


dberzano:
Currently being tested for:

* `build/O2/macos`
* `build/O2/Common/macos`

Let's see if they work properly before proceeding.

PR #574, opened 2018-06-12
Build all Control RPMs

teo:


dberzano:
Have you checked whether all our dependencies have Modulefiles?

teo:
@dberzano all the Control-related packages pulled in by control.sh indeed have Modulefiles, and so do their dependencies. Are there others I should check?

dberzano:
You can see here the dependency graph of the Control package:

graph

Do all the listed packages (in green or red) have modulefiles? And, for the record, red packages are those who are build requirements for some packages, and runtime requirements for others, indicating that there's probably (but not necessarily) a problem in the way dependencies are specified.

Notably:

* Control should not require `golang` as _runtime_ dependency since it's already available as a _build_ dependency for other pacakges - unless you want to have `go` available for further development, in this case please ignore this remark.
* coconut should probably depend on `protobuf` at runtime too.

p.s. the functionality for generating this is in alisw/alibuild#542

teo:
* I would indeed like to have `go` available for further development, as using upstream binaries via aliBuild is the best way I currently have to get the latest Go toolchain on CC7. Whether this happens always when loading the Control modulefile, or only when loading the golang modulefile manually, I don't really care and I'll proceed in whichever way you instruct.
* coconut only needs protobuf at build time, and even then it only uses the `protoc` binary and doesn't link against anything. `protoc` in turn talks to a native Go generator plugin (gogoproto), but this is handled by Control's Makefile. Either way, the resulting binary is static with no need for libprotobuf.

dberzano:
I confirm that everything is fine.

PR #578, opened 2018-06-21
Publish JAliEn on prod and improve tests

dberzano:


dberzano:
@costing @mimarpe this means you will be notified via email when a new JAliEn release hits CVMFS.

PR #584, opened 2018-07-30
Publish pythia6 on prod and ROOT6 on nightlies

dberzano:


dberzano:
@qgp you'll receive an automatic email when published!

PR #585, opened 2018-08-02
Publish: exclude packages older than X days

dberzano:
This helps preventing publishing too many packages at the same time due to a simple configuration mistake. It also assumes that "everything was fine" during the last X days and there is no "hole" (missing packages) to fill, speeding up publishing considerably.

Configure it to a reasonable value: by default the old behaviour is unchanged (never exclude anything)


dberzano:
@teo I wanted to implement this long awaited feature together with the `DATAFLOW` fix. Once merged, your package will be generated!

PR #589, opened 2018-08-08
Install aliBuild with pip at all times

dberzano:
Ensures Python dependencies are correctly installed as well. Applies to:

* Continuous integration script
* DA RPMs generation

We also clean up unused scripts.


ktf:
How do we test unreleased aliBuild versions now?

ktf:
ah, ok. pip magic. fine with me.

dberzano:
Yes, you can `pip install` from git, that's the trick 😄 

This method is already used pretty much everywhere, those only two scripts were missing AFAIK.

PR #604, opened 2018-09-26
Use distinct build names across different CI jobs

dberzano:
See O2-373


ktf:
How come they clash in the first place? Creation of packages should be atomic and parallel builds should work just fine. Maybe this is an issue in alibuild?

dberzano:
As discussed privately, the explanation's in the ticket [O2-373](https://alice.its.cern.ch/jira/browse/O2-373) - it's indeed an aliBuild bug as well, but I have no idea how to solve it on the aliBuild side quickly and without breaking other stuff 😞

ktf:
Ok, weird.

PR #620, opened 2018-11-08
Install many utility scripts with pip

dberzano:
It has the clear advantage that dependencies are installed correctly as well.
This is a solution to the ugly workaround in #618. The solution is also extended
to other components.


brinick:
Hi @dberzano,
Should the "--ignore-install" option instead be "--ignore-installed"?

dberzano:
It is the same. Shortcuts are recognized.

brinick:
Ah yes, forgot about that one 😫

PR #621, opened 2018-11-08
Install PR permissions checker with pip

dberzano:


dberzano:
Tested, and also makes @brinick happy by using the more correct `pip` switch 🙂

PR #625, opened 2018-12-04
Set CI environment to UTF-8

dberzano:


dberzano:
@Barthelemy fyi...

ktf:
Seriously? :-)

dberzano:
Oh yes. This surprised me. Here's a nice reproducer:

```python
# encoding: utf-8
from __future__ import unicode_literals
import os
import shutil
os.mkdir('temp')
with open('temp/привет.html'.encode('utf-8'), 'w'):
    pass
print(os.listdir(b'temp')[0])
shutil.rmtree('temp')
```

Call it `blah.py`. With Python 3, no problem at all (of course). With Python 2, the following will fail:

```bash
LC_ALL="C" python2 blah.py
```

and the following will work:

```bash
LC_ALL="en_US.UTF-8" python2 blah.py
```

PR #631, opened 2019-01-25
Generate RPMs for aliECS GUI

awegrzyn:


awegrzyn:
@dberzano

dberzano:
@awegrzyn note that there is a filter in the publisher that ignores packages produced more than 7 days ago for performance reasons. You may want to retag and rebuild the package to trigger all RPM generations from now on.

PR #633, opened 2019-02-07
Run guntest on macOS tests for AliRoot

dberzano:


dberzano:
@pzhristov fyi. This is in production since a while, but was never committed.

PR #651, opened 2019-05-08
Publish GEANT4 required by ALIROOT-8253

ktf:


ktf:
@dberzano can you please cross check?

PR #652, opened 2019-05-11
Fix to avoid duplicate logs.

aphecetche:
As far as I can tell all the text logs available from the PR checks (in ali-ci.cern.ch/repo/logs/AliceO2Group/AliceO2 at least) are duplicated (because of "latest" links).
This should fix it.

### DIFF
--- a/report-pr-errors
+++ b/report-pr-errors
@@ -116,7 +116,7 @@ class Logs(object):
             globbed.extend(glob(sp))
 
         suffix = ("latest" + "-" + self.develPrefix).strip("-")
-        logs = { x for x in globbed if dirname(x).endswith(suffix) }
+        logs = {x for x in globbed if dirname(x).endswith(suffix)}
 
         print("Found:\n%s" % "\n".join(logs), file=sys.stderr)
         self.logs = logs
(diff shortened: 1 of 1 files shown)


aphecetche:
well, my python is pretty young, but I thought this was a set comprehension and thus was working not by accident ? But, ok, I'm not 100% at ease with generator and comprehension(s) (French pun might be intended ...)
Are you positively sure it's wrong ?

ktf:
@aphecetche you are right:

```python
>>> type({x for x in [1]})
<class 'set'>
type({x:1 for x in [1]})
>>> type({x:1 for x in [1]})
<class 'dict'>
```

report-pr-errors
```diff
@@ -116,7 +116,7 @@ class Logs(object):
             globbed.extend(glob(sp))
 
         suffix = ("latest" + "-" + self.develPrefix).strip("-")
-        logs = [x for x in globbed if dirname(x).endswith(suffix)]
+        logs = { x for x in globbed if dirname(x).endswith(suffix) }
```
ktf: why the extra spaces?

aphecetche: A collateral damage of an unwanted removal of a virtualenv that contained my python formatter ;-)  Now fixed.

PR #658, opened 2019-06-28
Publish `NEXT_ALIEN` releases for JAlien commissioning

ktf:


ktf:
@atlantic777 @MichaelLettrich

ktf:
Given if I understand correctly this will not be tested until monday at this point, any objections to postpone it to then?

Atlantic777:
@ktf in fact, I believe we can test it before Monday (@mgheata). Any chance to get this today?

PR #668, opened 2019-08-05
publish Readout RPMs

sy-c:


dberzano:
Hold on, let me verify whether we really want to produce RPMs for _all_ builds of Readout, this may diverge quickly.

sy-c:
Indeed, I have not taken into account such subtlety ! What I would like is just that when there is a Jenkins job triggered manually for PACKAGE_NAME=Readout I get the corresponding RPM(s) for the version(s) as defined in OVERRIDE_TAGS

dberzano:
I have commented on Jira :-)

PR #673, opened 2019-08-10
mesos-dns-lookup: improve reliability

ktf:
* Error handling in case of bad arguments.
* Provide sensible defaults for MESOS_DNS variable.
* In case we find no result with the web API, return the provided
  argument as fallback as the local DNS might still provide a viable
  answer (e.g. if a local caching dnsmasq is provided).


ktf:
After discussing with @pzhristov we merge this and we will take care of any fallout which might result from it separately.

PR #674, opened 2019-08-10
Use actual hostnames rather than IPs

ktf:


ktf:
@dberzano could you remind me why you needed the actual IPs and not the hostnames?
@MichaelLettrich FYI.

ktf:
After discussing with @pzhristov we decided to merge this and eventually take care of the issue deriving from it (e.g. where CERN DNS is not available) separately.

PR #698, opened 2019-09-02
report-pr-errors: force inline logs

ktf:


ktf:
@davidrohr this should fix the "download logs" issue.

PR #717, opened 2019-10-08
Publish alibuild logs

ktf:


ci/continuous-builder.sh
```diff
@@ -341,7 +341,8 @@ while true; do
     [[ $BUILD_IDENTIFIER ]] || BUILD_IDENTIFIER=${CHECK_NAME//\//_}
 
     FETCH_REPOS="$(aliBuild build --help | grep fetch-repos || true)"
-    ALIBUILD_HEAD_HASH=$pr_hash ALIBUILD_BASE_HASH=$base_hash                    \
+    rm -fr aliBuild.log
```
teo: Is this still ok if there's no old log file?

ktf: Yes:

```
→ rm foo
rm: cannot remove 'foo': No such file or directory
→ echo $?
1
→ rm -f foo
→ echo $?
0
```

PR #718, opened 2019-10-09
RPM publishing: address O2-852, O2-437, O2-849 and O2-853

ktf:
* Add test package to those being published.
* Exclude compile_commands.json from RPMS, so that they do not
  conflict in the updatable RPMS.

### DIFF
--- a/publish/aliPublish-rpms.conf
+++ b/publish/aliPublish-rpms.conf
@@ -10,6 +10,7 @@ architectures:
   slc7_x86-64:
     RPM: el7.x86_64
     include:
+      rpm-test: True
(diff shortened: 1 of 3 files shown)


ktf:
@teo I think I've things under control now, this should also remove `compile_commands.json` from the rpms. Can you give it a spin, again?

teo:
> @teo I think I've things under control now, this should also remove `compile_commands.json` from the rpms. Can you give it a spin, again?

Thank you @ktf . I'm on official travel right now so it's a bit difficult, but I'll definitely give it a spin when I get back next week.

PR #721, opened 2019-12-09
[WIP] macOS Catalina + launchd

MichaelLettrich:
Changes to continuous builder scripts to allow migration of CI to macOS Catalina

* read only root file system requires moving `/build` to `/home/alibuild/build`
* `launchd` scripts to run continuous builders and log rotation.
* changes in `run-continous-builder.sh` to adapt to being used by `launchd`


MichaelLettrich:
DO NOT MERGE THIS. an accidental pull on Mojave machines of these updates will cause lots of angry mails.

PR #722, opened 2019-12-12
O2 946 feature macos catalina

MichaelLettrich:
Incorperate minimal changes for macOS Catalina in run-continous-builder. This will break compatibility with current mac builders as it changes the location of `/build` to `/usr/local/build`


ci/run-continuous-builder.sh
```diff
@@ -61,7 +61,7 @@ export MONALISA_PORT=8885
 export MAX_DIFF_SIZE=20000000
 export DELAY=20
 export DEBUG=true
-export MIRROR=/build/mirror
+export MIRROR=/usr/local/build/mirror
```
ktf: No, you can create /build in catalina.

ktf: man synthetic.conf

PR #742, opened 2020-03-19
List PR for testing if there is at list one approved review

ktf:


ktf:
@jgrosseo

ktf:
This should allow testing PRs of newcomers to AliceO2, if at least one review is present.

PR #748, opened 2020-04-22
Report PR timing to Google Analytics

ktf:
Plan is to deprecate and remove our own InfluxDB / Grafana instance
as we do not have manpower to maintain it.


ktf:
Will try to deploy this after lunch.

PR #749, opened 2020-04-22
Rework continuous builder (O2-1366)

ktf:


ktf:
Currently being tested for the Centos8 CI.

PR #750, opened 2020-04-23
Do not hardcode updatable RPMs version (O2-1360)

ktf:


ktf:
@awegrzyn @Barthelemy I assume this should be the wanted behaviour, correct? Do you see any case were this would not be the case for the updatable RPMs?

awegrzyn:
@ktf After quick look it's fine by me, but I think we shall discuss it within FLP project

ktf:
Ok. Let me know.

ktf:
Ping?

awegrzyn:
Hi, we have scheduled meeting tomorrow at 10 to discuss that

awegrzyn:
@ktf It's good to go

awegrzyn:
@ktf I recreated RPMs yesterday, but they still use `=` instead of `>=`

ktf:
Mmmm... I might have to redeploy. :-/ Let me check.

ktf:
Yes, sorry about that. Can you try again?

PR #760, opened 2020-05-12
Simplify build-any-ib

ktf:
* Less gymnastic with alibuild installation
* Avoid filling the out of bound "/build" folder, now that we
  use static workers.


ktf:
@awegrzyn @graduta this will hopefully fix your issues with the full disks.

PR #762, opened 2020-05-14
report-pr-errors: add date to the logs

ktf:


ktf:
@MichaelLettrich this should do it. We then need to update all the builders to pick up this version of report-pr-errors, however that does not require fixes.

PR #771, opened 2020-06-24
DO NOT MERGE Add symlink for O2/latest

ktf:


ktf:
@Atlantic777: will this (i.e. the same file which keeps changing) be a problem for CVMFS ?
@costing @jgrosseo: will this work? what happens if people start to ask for trains with O2/latest? 
Suggested / requested by @sawenzel.

ktf:
To be clear, my worry is that O2/latest is resolved late and part of a train uses a version / part a different one.

jgrosseo:
This should not enter the package table. We cannot bookkeep and clone trains anymore when the version string is not uniquely pointing somewhere.

PR #773, opened 2020-06-26
cvmfs-monitor: follow up on cvmfs monitor API change

janpriessnitz:
CVMFS monitor service API has changed. Need to change
ALICE cvmfs monitoring script to stay compatible.

### DIFF
--- a/cvmfs-mon/cvmfs-mon.yml
+++ b/cvmfs-mon/cvmfs-mon.yml
@@ -1,45 +1,33 @@
 ---
 repos:
-  alice-ocdb.cern.ch:
-    CERN:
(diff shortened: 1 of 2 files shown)


cvmfs-mon/cvmfs-mon.yml
```diff
@@ -49,7 +37,7 @@ notif:
 
     Summary:
 
-      http://cernvm-monitor.cern.ch/cvmfs-monitor/%(repo)s/
+      https://cvmfs-monitor-frontend.web.cern.ch/%(repo)s/
```
jblomer, line 40: This can probably remain the old URL because we'll to migrate the name.

PR #785, opened 2020-07-28
Whitelist O2 GSI specific test builds

Atlantic777:
This PR should enable publishing of individual O2 builds with JAliEn-ROOT and TNetXNG file access backend to be tested at GSI. There might be more GSI specific test builds in the future so `_GSI` tag should be ok for this purpose.

cc @costing


ktf:
I was actually thinking that we should simply remove the whitelisting and simply keep an exclusion list (if really needed), especially given that recent developments allow us to build reusing CVMFS installation, so we need to publish also the build requires...

PR #788, opened 2020-08-24
GPU builder with slc8 no longer needs --always-prefer-system

davidrohr:


ktf:
Does this mean we should warmup moving to CC8?

davidrohr:
Yes, I just created a new docker image for the gpu builder using slc8, cuda11, and rocm 3.7

PR #798, opened 2020-10-07
Add RPM cleanup script

TimoWilken:
This script tries to detect RPMs that aren't required by key packages any more,
so that they can be deleted.

I've only tested this script with some, not all, RPMs and not implemented e.g.
fully general version matching. Any untested or future RPMs may well break this
script.

Note that nothing calls this script so far!

### DIFF
--- a/publish/aliPublish.conf
+++ b/publish/aliPublish.conf
@@ -116,6 +116,11 @@ architectures:
     AliEn: el6-x86_64
     RPM: false
     include:
+      AliDPG:
+       # Old convention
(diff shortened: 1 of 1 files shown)


TimoWilken:
I was changing other stuff in my fork of ali-bot so I made these commits into a separate branch. I didn't realise that broke this PR!

TimoWilken:
I'll make a new PR. Pushing the new commits onto my fork's master branch doesn't seem to fix this PR.

PR #806, opened 2020-10-21
Only print untested PRs

TimoWilken:
Don't fall back to other PRs, to avoid slowing down the build.

Without this change, the CI builder will wait for some minutes and test
already-tested PRs again before moving on to the next repo, which may have
untested PRs.

### DIFF
--- a/list-branch-pr
+++ b/list-branch-pr
@@ -245,6 +245,12 @@ def parseArgs():
                         default="",
                         help="Execute a script on the resulting PR")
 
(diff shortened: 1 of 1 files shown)


ktf:
Could you made this optional (i.e. via command line option)? so that we can bring back the old behaviour, if needed.

TimoWilken:
I've added a `--fallback-to-untested` option to bring back the old behaviour.

PR #812, opened 2020-10-23
Improvements to CI builder

TimoWilken:
This PR:

- exec's continuous-builder.sh instead of looping, so updates to that script are picked up
- sleeps if there are no PRs to build in any repository, so GitHub API calls aren't used up too quickly
- gets rid of a pointless loop in list-branch-pr
- uses a common GitHub API cache for list-branch-pr

### DIFF
--- a/list-branch-pr
+++ b/list-branch-pr
@@ -110,6 +110,7 @@ def process_pull(pull, cgh, args):
         print("Processing: %s" % pn, file=sys.stderr)
         try:
             item = _do_process_pull(pull, cgh, args)
+            print(item, file=sys.stderr)
             print("Processing: %s. Done." % pn, file=sys.stderr)
         except RuntimeError as e:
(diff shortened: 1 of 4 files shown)


list-branch-pr
```diff
@@ -305,25 +306,23 @@ def send(script, pulls):
 if __name__ == "__main__":
     args = parseArgs()
 
-    cache = PickledCache(".cached_github_client_cache")
+    cache = PickledCache("../.cached_github_client_cache")
```
ktf, line 309: It would probably better to make this configurable or use $HOME. Otherwise it might fail for local tests.

TimoWilken: Good point, I'll make it `$HOME` in a separate PR.

TimoWilken: See #814

PR #814, opened 2020-10-23
Store GitHub cache in $HOME

TimoWilken:
This shares the cache between difference repositories in the unified CI builder (saving API requests), but doesn't rely on a specific directory structure (and therefore supports local testing more easily).

### DIFF
--- a/list-branch-pr
+++ b/list-branch-pr
@@ -2,6 +2,7 @@
 from __future__ import print_function
 
 import functools
+import os.path
 import random
 import sys
 import time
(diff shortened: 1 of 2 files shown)


list-branch-pr
```diff
@@ -306,7 +307,7 @@ def send(script, pulls):
 if __name__ == "__main__":
     args = parseArgs()
 
-    cache = PickledCache("../.cached_github_client_cache")
+    cache = PickledCache(os.path.expanduser("~/.cached_github_client_cache"))
```
ktf, line 310: How about making it configurable, defaulting to $HOME?

PR #815, opened 2020-10-24
Delete o2checkcode.env

ktf:


ktf:
@davidrohr are you happy with the GPU builder taking over this?

ktf:
Dropping the extra check for now. @TimoWilken on monday we should rename the gpu test to "full-ci".

PR #822, opened 2020-10-26
Rename GPU build to full CI

TimoWilken:


davidrohr (approved):

ktf:
You need to change the "required tests" in the github settings as well, IIRC.

TimoWilken:
@ktf I can't seem to add a custom check in that list, it only lets me choose from builds that have run in the last week - so maybe the renamed build has to run once, then it can be selected? Or I could use `report-pr-errors` to add a bogus result to some old commit, maybe it'll let me choose it then.

ktf:
Yes, I think it needs to run once. You can use `ali-bot/set-github-status` to set the fake test.

TimoWilken:
@ktf, done! I've changed the branch settings for `dev` in `AliceO2` such that `build/O2/fullCI` is required and `build/O2/gpu` isn't.

PR #824, opened 2020-10-26
Extract error messages into separate HTML file

TimoWilken:
This extracts the important messages out of the large log files, so users can find the more easily. The HTML file is uploaded to S3 alongside the full log file, and links to the latter so both are accessible. The GitHub status will link to the HTML file directly.

Untested so far; I'll work out how to do that sensibly tomorrow.

### DIFF
--- a/report-pr-errors
+++ b/report-pr-errors
@@ -121,13 +121,15 @@ class Logs(object):
         self.full_log_latest = self.constructFullLogName(args.pr, latest=True)
         self.pretty_log = self.constructFullLogName(args.pr, pretty=True)
         self.dest = args.logsDest
-        self.url = join(args.logsUrl, self.constructFullLogName(args.pr, pretty=True))
+        self.url = join(args.logsUrl, self.pretty_log)
+        self.log_url = join(args.logsUrl, self.full_log)
+        self.build_successful = args.success
 
     def parse(self):
         self.find()
         self.grep()
-        self.generate_pretty_log()
         self.cat(self.full_log)
+        self.generate_pretty_log()
         if self.is_branch:
             self.cat(self.full_log_latest, no_delete=True)
         if self.dest.startswith("rsync:"):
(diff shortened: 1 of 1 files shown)


davidrohr:
WIthout trying it out, some comments:
- For the o2checkcode recipe: we should not show all errors but only those that will come in the list after "List of errors". There is already some grepping before. Just grepping for `error:` might yield false positives from system includes.
- Could we also check for failed ctests, I think the output is something with `... failed` not with `error:`, and possibly list the name of the failed tests.
- There should be some comment if the build fails but the grep does not extract any error, hinting that it is an uncommon error and the full log is to be inspected.

davidrohr:
Is there a simple way I could see how the page would look like on a real PR?
That would help me to judge what could be improved :)

TimoWilken:
Thanks for the feedback, @davidrohr!

I've improved the tool by extracting the error list from `o2checkcode`, grepping for failed ctests, and adding a few explanatory messages. I've kept the default error log as it might be more useful than just the raw log (which is linked to), but I can remove it easily if there are any objections.

For an example of the output, see: <https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/Configuration/65/aa941fdd22f15bcf99f5548574b7b7104f2422a5/build_Configuration_o2-dataflow/pretty.html> -- note however that I've since made the tool clean its output up a bit more, showing messages if nothing was found for `o2checkcode` or unit tests.

davidrohr (approved):
ok, thx, looks reasonable at first glance, but it would be helpful to see how it looks on PRs that have failed, in particular that have failed for different reasons.

But in any case, you have a link to the full log in there. Just to be sure, the full log is like before, so it can be the combination of multiple logs when there were multiple logs before? E.g., if the O2Suite build for an O2 PR failed at QC, the log would show both the O2 and the QC build log.

If that is still the case, from my side we can merge this, it is at least as good as before, then we will see quickly how it behaves and can improve incrementally.

TimoWilken:
@davidrohr, do you happen to have a PR handy that failed `build/O2/o2`? Otherwise I'll try and find one to test.

I haven't changed the way the full log is generated, so it's the same as before! It combines all the logs can find, so as far as I know it'll show both the O2 and QC logs in that case.

PR #833, opened 2020-10-30
Rework CI builder for multiple repositories

TimoWilken:
`list-branch-pr` now returns PRs to build from multiple repositories, making PR discovery faster. This also lets the CI builder give better feedback to queued PRs and enables proper fallback to previously-tested PRs in the same builder when there are no untested PRs available.

Instead of looping through repositories and then PRs within those, the outer loop is now done by `list-branch-pr` during its search for PRs. This gets rid of a lot of repetitive pointless actions (like reinstalling ali-bot many times) during PR discovery. The 'inner loop' now simply processes the PRs returned by `list-branch-pr` directly, without having to loop through repos.

A structural change is that `list-branch-pr` now parses `*.env` files to determine which repositories it should query for PRs. This seems to work pretty well so far, but it means that `*.env` files should pretty much stay as simple key-value stores without using too many shell features or calling external commands. I changed the format of the variables only used by list-branch-pr (#% directives) to make it obvious that they are slightly magic and not used by the shell scripts.

I also removed unused options from `list-branch-pr`, and added fallbacks to environment variables for the options where this is useful.

Finally, I improved the hashing algorithm used for deciding whether to build a PR -- the check name is now included in the hash, so work for one PR is spread among multiple workers.

Some tweaking might still be required:

- How many PRs should be built at a time? Longer queues mean more feedback for PRs further down the queue but potentially longer delays until new PRs are picked up.


TimoWilken:
@ktf these changes are tested (on `mesosdaq/slc7`) and ready to be merged.

ktf:
Very nice, thanks.

Regarding the .env files remaining simple KV pairs, I think it's a good idea in any case.

Regarding how often to check for already processed PRs, I would say the logic should be:

* Process all the non-tested PRs, then query again.
* Process one of the red PRs, then query again.
* Process one of the green PRs then query again.

ci/repo-config/mesosci/slc7/o2-dataflow.env
```diff
@@ -5,7 +5,7 @@ ALIBUILD_DEFAULTS=o2-dataflow
 PR_REPO=alisw/alidist
 PR_BRANCH=master
 NO_ASSUME_CONSISTENT_EXTERNALS=
-TRUST_COLLABORATORS=true
+#% TRUST_COLLABORATORS true
```
ktf, line 8: What is this for?

TimoWilken: Only `list-branch-pr` uses this, so I thought this would make it clear that `TRUST_COLLABORATORS` (and `TRUSTED_USERS`) is magic, to stop people from removing the variables because they aren't used by the shell scripts.

TimoWilken:
> Regarding the .env files remaining simple KV pairs, I think it's a good idea in any case.

The only weirdness is [`mesosci/slc8-gpu/o2-fullci.env`](https://github.com/alisw/ali-bot/blob/d40885676f502c1ae63a2555d308a9439dbd6296/ci/repo-config/mesosci/slc8-gpu/o2-fullci.env), which sources `/entrypoint.sh` in the docker container, which contains:

```bash
#!/bin/bash
CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}:/opt/rocm/lib/cmake
AMDAPPSDKROOT=/opt/amd-app
ALIBUILD_O2_FORCE_GPU=1
PATH="${PATH}:/opt/cuda/bin"
export CMAKE_PREFIX_PATH AMDAPPSDKROOT ALIBUILD_O2_FORCE_GPU PATH
if [ ! "X$@" = X ]; then
  exec "$@"
fi
```

If nothing else uses this file, maybe it's best to merge it into `o2-fullci.env`?

> Regarding how often to check for already processed PRs, I would say the logic should be:
> 
>     * Process all the non-tested PRs, then query again.
>     * Process one of the red PRs, then query again.
>     * Process one of the green PRs then query again.

OK, I'll change the logic to match that!

TimoWilken:
Regarding the discussion about reserving capacity for rebuilding PRs even if untested PRs are available, should we do that? If yes, the easiest way might be to have an X% chance on each invocation of `list-branch-pr` of skipping tested PRs and just returning a red (or green, if no red available) PR instead.

Also, should there be a similar mechanism of returning green PRs sometimes, even if red ones are available? As it is currently, a single broken PR (that keeps failing the build) will completely prevent any green PRs from being rebuilt.

ci/repo-config/mesosci/slc7/o2-dataflow.env
TimoWilken: @ktf Would you prefer it if I made those shell variables (i.e. `TRUST_COLLABORATORS=true` as before)?

PR #844, opened 2020-11-09
Make testing different ali-bot and alibuild versions easier

TimoWilken:
With this patch, the ali-bot and alibuild versions to install can be controlled on a per-worker basis. This makes it easier to test patches to those on a single worker before deploying.


TimoWilken:
I've rebased it as it conflicted with #845 -- fine to merge now, @ktf.

PR #845, opened 2020-11-10
Remove run-continuous-builder.sh wrapper

TimoWilken:
The wrapper mostly does the same setup as continuous-builder.sh; what it did extra before is now incorporated into continuous-builder.sh as well.

MacOS builds are now handled by continuous-builder.sh.

A few more variables need to be initialised outside the script on MacOS.


ci/repo-config/macos/macos/DEFAULTS.env
```diff
@@ -0,0 +1,5 @@
+NO_ASSUME_CONSISTENT_EXTERNALS=true
+TRUSTED_USERS=ktf,dberzano
```
ktf, line 2: Please drop dberzano and add yourself.

ci/continuous-builder.sh
```diff
@@ -3,45 +3,79 @@
...
+      export "${var?}"
+    fi
+  done
+
+  case "$(uname)" in
+    # The MacOS builders aren't in the right directory at first, apparently.
+    Darwin) ci_work_dir=/build/ci_checks/alice_ci_$WORKER_INDEX
+            mkdir -p "$ci_work_dir"
+            cd "$ci_work_dir" || exit 1;;
+  esac
```
ktf, line 23: This should be solved by how we start it on mac, I would rather avoid extra stuff in here for something that is really usage specific.

ci/continuous-builder.sh
```diff
@@ -84,4 +130,10 @@ fi
 # Get updates to ali-bot
 TIMEOUT=$(get_config_value timeout "${TIMEOUT:-600}") reset_git_repository ali-bot
 
-exec continuous-builder.sh
+# Run CI builder under screen if possible and we're not already there. This is
+# for the macOS builders.
+if [ -n "$STY" ] || ! type screen > /dev/null; then
+  exec "$0" --skip-setup "$@"
+else
+  exec screen -dmS "ci_$WORKER_INDEX" "$0" --skip-setup "$@"
```
ktf, line 138: I never liked the idea of running under screen. Can you check how we can make sure we run under systemd?

ci/continuous-builder.sh
```diff
@@ -15,12 +15,9 @@ if [ "$1" != --skip-setup ]; then
     fi
   done
 
-  case "$(uname)" in
-    # The MacOS builders aren't in the right directory at first, apparently.
-    Darwin) ci_work_dir=/build/ci_checks/alice_ci_$WORKER_INDEX
-            mkdir -p "$ci_work_dir"
-            cd "$ci_work_dir" || exit 1;;
-  esac
+  if [ -r ~/.continuous-builder.sh ]; then
```
ktf, line 18: I would simply drop it. My preferred solution is that we either run aurora on mac as well (we might have to update our own mesos quite a bit though) or that we use launchd.

ci/continuous-builder.sh
TimoWilken: @MichaelLettrich could you send me the old launchd scripts? I'll see what I can do with those.

@ktf it might be better to keep screen for now so the builds work again, then I can migrate to launchd some time this week when there's less time pressure.

ci/continuous-builder.sh
TimoWilken: Right. For now, that file apparently contains secrets that we need. As above, is it OK if we keep the old way for now so the builds work again?

ktf: agreed.

TimoWilken:
@MichaelLettrich:

`~/.continuous-builder.sh` needs to have these additional variables:

```bash
MESOS_ROLE=macos
CONTAINER_IMAGE=(macos1, macos2 or macos3 depending on the machine)
```

...plus the constants defined [here](https://gitlab.cern.ch/ALICEDevOps/ali-marathon/-/blob/master/aurora/continuous-integration.aurora#L20): `GITHUB_TOKEN`, `GITLAB_USER`, `GITLAB_PASS`, `CODECOV_TOKEN`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `INFLUXDB_WRITE_URL`, `ALIBOT_ANALYTICS_ID`, `MONALISA_HOST`, `MONALISA_PORT`, `WORKER_INDEX`, `WORKERS_POOL_SIZE`.

Currently alidist/O2Suite, O2, AliRoot, Common and QualityControl are being built. How should they be distributed among the 3 machines?

MichaelLettrich:
@TimoWilken see https://alice.its.cern.ch/jira/browse/O2-1817 .
would it be possible to have:
`alibuildmac00`, `alibuildmac01`, `alibuildmac03` as container  image names so that the names map 1:1? (there currently is no `alibuildmac02`, but it should arrive soon)

TimoWilken:
@MichaelLettrich done! Now you just need `CONTAINER_IMAGE=$(hostname)` in `~/.continuous-builder`.

Note that I changed the name of that config file back to what it was before (removing the accidental `.sh`).

Could you also add the following mac-specific code to each `~/.continuous-builder` (or execute this before invoking `continuous-builder.sh`):
```bash
ci_work_dir=/build/ci_checks/alice_ci_$WORKER_INDEX
mkdir -p "$ci_work_dir"
cd "$ci_work_dir" || exit 1
```

MichaelLettrich (approved):
Thanks for the great work. This looks like it will work fine for the macOS machines. I think I can go and deploy this with your help.

TimoWilken:
@MichaelLettrich great, thanks!

`~/.continuous-builder` should contain the constants defined [here](https://gitlab.cern.ch/ALICEDevOps/ali-marathon/-/blob/master/aurora/continuous-integration.aurora#L20): `GITHUB_TOKEN`, `GITLAB_USER`, `GITLAB_PASS`, `CODECOV_TOKEN`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `INFLUXDB_WRITE_URL`, `ALIBOT_ANALYTICS_ID`, `MONALISA_HOST`, `MONALISA_PORT`, `WORKER_INDEX`, `WORKERS_POOL_SIZE`, plus:

```bash
MESOS_ROLE=macos
CONTAINER_IMAGE=$(hostname --short)
ci_work_dir=/build/ci_checks/alice_ci_$WORKER_INDEX
mkdir -p "$ci_work_dir"
cd "$ci_work_dir" || exit 1
```

Then just run `continuous-builder.sh`.

@ktf, can we merge this PR so it's simpler to deploy?

PR #854, opened 2020-11-12
Don't hardcode pip version

TimoWilken:
According to <https://alisw.github.io/infrastructure-macos>, we use `pip3` on MacOS. The `pip` binary does the right thing on Linux and MacOS in this case, selecting the right version.

### DIFF
--- a/ci/build-helpers.sh
+++ b/ci/build-helpers.sh
@@ -58,7 +58,7 @@ function pipinst () {
 
   # Sometimes pip gets stuck when cloning the ali-bot or alibuild repos. In
   # that case: time out, skip and try again later.
-  short_timeout pip2 install $pip_user --upgrade --upgrade-strategy only-if-needed "git+https://github.com/$1"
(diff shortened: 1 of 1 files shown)


ktf:
I am not sure this is actually the case on SLC6.

TimoWilken:
Do we build on SLC6? I thought it was just 7 and 8. All the docker containers the CI builders use have `pip --version` == `pip 20.2.4 from /usr/lib/python2.7/site-packages/pip (python 2.7)`.

ktf:
Releases in jenkins are actually built on SLC6, so if we plan to use this to do the nightly builds, we might have to.

ktf:
That said, we can probably think about it when we get to it...

TimoWilken:
Otherwise, putting:
```bash
pip2 () { pip "$@" }
```
into `~/.continuous-builder` for the MacOS workers might work for now.

PR #861, opened 2020-11-13
Remove old workarounds

TimoWilken:
These are left-over workarounds from the MacOS builder that don't apply any more.

### DIFF
--- a/ci/build-loop.sh
+++ b/ci/build-loop.sh
@@ -159,7 +159,7 @@ FETCH_REPOS="$(aliBuild build --help | grep fetch-repos || true)"
 
 if ALIBUILD_HEAD_HASH=$PR_HASH ALIBUILD_BASE_HASH=$base_hash             \
                      clean_env long_timeout aliBuild                     \
-                     -j "${JOBS:-$JOBS_DEFAULT}" -z "$BUILD_IDENTIFIER"  \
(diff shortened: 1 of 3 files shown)


MichaelLettrich (approved):
OK as discussed,.

ci/build-loop.sh
```diff
@@ -159,7 +159,7 @@ FETCH_REPOS="$(aliBuild build --help | grep fetch-repos || true)"
 
 if ALIBUILD_HEAD_HASH=$PR_HASH ALIBUILD_BASE_HASH=$base_hash             \
                      clean_env long_timeout aliBuild                     \
-                     -j "${JOBS:-$JOBS_DEFAULT}" -z "$BUILD_IDENTIFIER"  \
+                     -j "${JOBS:-$(nproc)}" -z "$BUILD_IDENTIFIER"       \
```
ktf, line 162: I am not sure this is a good idea. a docker container will report as many cores as CPUs, but might be rate limited to less.

TimoWilken: Good point, though this is only the fallback -- at the moment, all checks have `JOBS` defined anyway, so it never gets to the fallback currently.

I could hardcode a number there (4 would work for the Linux and Mac builders, but might be a bit conservative) or I could make the `JOBS` variable mandatory instead so we don't need a fallback at all.

ktf: Fine as is.

PR #864, opened 2020-11-16
Software collection package to install all of O2 dependencies

ktf:


ktf:
This can be built by doing:

```bash
yum install -y rpm-build scl-utils-build
rpmbuild -ba o2-prereq.spec
```

PR #875, opened 2020-11-26
Simplify repo-cleanup/gen-repo-report.sh

ktf:


ktf:
@TimoWilken  could you review?

repo-cleanup/gen-repo-report.sh
```diff
@@ -1,60 +1,28 @@
...
-      find . -type f -printf "store $A %s %Cs %p\n" >> $T_REPORT
-    popd > /dev/null
-
-    pushd dist > /dev/null
-      find . -type l -printf "dist $A %p\n" >> $T_REPORT
-    popd > /dev/null
-
-  popd > /dev/null
+for A in `find $TARBALLS_PREFIX -maxdepth 1 -type d -printf "%f\n"`; do
+  mkdir -p $TARBALLS_PREFIX/$A/{store,dist}
```
TimoWilken, line 25: `{a,b}` is undefined in POSIX sh. It's probably easiest to just expand this manually.

repo-cleanup/gen-repo-report.sh
```diff
@@ -1,60 +1,28 @@
...
-    pushd store > /dev/null
-      find . -type f -printf "store $A %s %Cs %p\n" >> $T_REPORT
-    popd > /dev/null
-
-    pushd dist > /dev/null
-      find . -type l -printf "dist $A %p\n" >> $T_REPORT
-    popd > /dev/null
-
-  popd > /dev/null
+for A in `find $TARBALLS_PREFIX -maxdepth 1 -type d -printf "%f\n"`; do
```
TimoWilken, line 24:
```bash
cd "$TARBALLS_PREFIX"
for A in */; do ...
```
would match all directories directly under `$TARBALLS_PREFIX` and is less prone to whitespace-related trouble than this find expression.

Since everything in the loop happens in `$TARBALLS_PREFIX`, `cd`ing there first might make things a bit simpler.

TimoWilken (commented):
Two small comments -- other than that, looks good to me!

None of the variables are quoted, which might be a problem if there's whitespace in any of the paths, but I suppose for the repos as they currently are that's not a problem.

repo-cleanup/gen-repo-report.sh
ktf: No, I disagree here. Having the absolute path makes it easier to debug by cut & pasting the command from the log. However I added the protection for empty spaces everywhere. Can you please review again?

TimoWilken: If you're worried about whitespace, I'd turn the loop into
```bash
find "$TARBALLS_PREFIX" -maxdepth 1 -type d -printf '%f\n' | while read -r A; do
  ...
done
```
which handles spaces in filenames fine (but only breaks on newlines in filenames).

One minor thing is the `"%f\n"` -- I'd turn that into `'%f\n'` to be safe as double-quoting might turn `\n` into `n`, but I'm not 100% sure about that. With single-quoting, the backslash definitely gets passed to find, which turns it into a newline.

ktf: PR please. ;-)

TimoWilken: Done! #877 :)

PR #880, opened 2020-11-27
Tag alidist when tagging packages as well

TimoWilken:
This PR also refactors `daily-tags.sh`. Significant changes include:

- replacing custom workspace locking code with `mktemp`
- using `git show-ref` instead of parsing `git ls-remote` output, to avoid server round-trips each time
- using the S3 remote store by default

@ktf: which of the following features are used, and which can be removed?

- tagging behaviour is pretty complicated at the moment -- should `REMOVE_RC_BRANCH_FIRST` and/or `DO_NOT_CREATE_NEW_TAG` just be made the non-configurable default?
- patching the YAML seems pretty hacky -- is it worth adding a flag to aliBuild directly to override the current package's tag and version?

### DIFF
--- a/daily-tags.sh
+++ b/daily-tags.sh
@@ -2,7 +2,9 @@
 # -*- sh-basic-offset: 2 -*-
 
 # Check for required variables
-: "${ALIDIST_SLUG:=alisw/alidist@master}" "${PACKAGE_NAME:?}" "${AUTOTAG_PATTERN:?}" "${NODE_NAME:?}"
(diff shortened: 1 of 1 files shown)


ktf:
Please let's have one change at the time, starting with the one at the top.

TimoWilken:
@ktf see #881

ktf:
Was there any reason why this PR was not moved to production?

ktf:
@TimoWilken ping can we close this?

TimoWilken:
@ktf, yes, most of these improvements were in different PRs already.

PR #881, opened 2020-11-27
Replace custom locking code with mktemp

TimoWilken:


daily-tags.sh
```diff
@@ -117,22 +101,26 @@ EOF
 
 diff -rupN alidist/defaults-${DEFAULTS_LOWER}.sh.old alidist/defaults-${DEFAULTS_LOWER}.sh | cat
 
+# Select build directory in order to prevent conflicts and allow for cleanups.
+workarea=$(mktemp -dp "$PWD" daily-tags.XXXXXXXXXX)
```
ktf, line 105: -p does not work on mac.

TimoWilken: Fixed, I've dropped the `-p` and made `$PWD` part of the filename template.

PR #888, opened 2020-12-01
Use S3 repo by default in daily-tags.sh

TimoWilken:


ktf:
Before we can do this, we need to make sure we can publish to CVMFS from S3.

TimoWilken:
OK! How do we do that?

ktf:
Currently the publishing is happening via `ali-bot/publisher`. What it does it's to query ali-ci.cern.ch to get the list of packages to be published. We need to adapt from the JSON based package list which we have there to the flat one returned by S3.

TimoWilken:
For aliPublish changes, see: #921

ktf:
We need to rethink this before it can be merged. As we need to make sure that the old (cc7) repository is fully consistent in S3 and can be used.

PR #903, opened 2020-12-04
Drop external aliswmod command

ktf:


ktf:
@awegrzyn for whenever you confirm you do not need it anymore.

awegrzyn:
I'll notify all parties but it's being used a lot so we'll need some transition period

awegrzyn (commented):
I think we may also want to enforce `4.x` by specifying `environment-modules >= 4.0` in https://github.com/alisw/ali-bot/blob/1b9e8d6baf39581768cd30c89c5c573752ddabf4/publish/pub-rpms-template.sh#L35

ktf:
How do we go forward with this?

awegrzyn:
We've already migrated `aliswmod` to `module` in FLP Suite@dev, the idea was to merge this PR just before v0.15.0 release (so in case a v0.14.x release is needed it still contains `aliswmod`).
As the release is planned for Wednesday I'd merge it this Monday.

awegrzyn (approved):

PR #908, opened 2020-12-09
Use tag, not branch, for daily releases

TimoWilken:


ktf:
Wait, this will not work because the tag is there only at the end, no?

TimoWilken:
The tag is assigned on line 29, so before the script that uses it

ktf:
Ok, but then alibuild will try to build a tag which is not yet tagged, no?

PR #909, opened 2020-12-09
Update pub-rpms-template.sh

ktf:


ktf:
@awegrzyn this is one thing. I think we need to rework aliPublish as well, though, since I do not see defaults-o2-dataflow among the dependencies.

PR #910, opened 2020-12-09
Modify both defaults-%s.sh and %s.sh; build using branch before tagging

TimoWilken:


TimoWilken:
I've tested this PR on my own forks of alidist and AliceO2. This is the release it generates: <https://github.com/TimoWilken/alidist/releases/tag/O2-testing-O2-20201211> -- @ktf, is that acceptable?

Tested with the following vars:
```bash
ALIDIST_SLUG=TimoWilken/alidist@own-o2-fork-for-testing  \
  PACKAGE_NAME=O2                                        \
  DEFAULTS=o2                                            \
  AUTOTAG_PATTERN=testing-O2-%Y%m%d                      \
  NODE_NAME=twilkendesktop                               \
  ali-bot/daily-tags.sh
```

ktf:
I think it's fine, but do we really need the reshuffling of the YAML?

TimoWilken:
@ktf, is it acceptable if only o2.sh is changed, not defaults-o2.sh?

The former would be fairly simple to patch using `sed`, but the latter requires adding keys or editing in just the right place that would be very annoying to do using `sed`/`awk`.

ktf:
Ciao,

As long as the default does not override the version of O2, I think it's
more than reasonable. We must make sure, however, that such a tag lives on
a branch (say v20.49-patches).

Ciao,
Giulio

TimoWilken:
Hm, some of the `defaults-*.sh` files override the version of the package they're supposed to build (e.g. next-root6 for the Ali* packages). The following awk script modifies the `defaults-*.sh` without rearranging, but it's a bit longer and more brittle than the Python script. I'll keep the Python script, if that's alright.

```awk
BEGIN { in_overrides = 0 }
# Print every line by default
{ printline = 1 }

# Detect end of overrides block
(/^[^ ]/) { in_overrides = 0 }
# Detect start of overrides block
/^overrides:$/ { in_overrides = 1 }

# Detect the override block for the current package
(in_overrides && /^  [A-Za-z].*:$/) {
  cur_pkg = $1
  sub(/:$/, "", cur_pkg)
}

# If this is an override for the package we are building, change overrides
(in_overrides && cur_pkg == PKG && /^    version: /) {
  print "    version: \"" VERSION "\""
  printline = 0
}
(in_overrides && cur_pkg == PKG && /^    tag: /) {
  print "    tag: \"" TAG "\""
  printline = 0
}

# Print other lines unmodified
(printline) { print }
```

TimoWilken:
I've retested this PR. This is the generated release: <https://github.com/TimoWilken/alidist/releases/tag/O2-testing-O2-20201211>. `defaults-o2.sh` is still a bit of a mess (PyYAML doesn't remember the original order of the keys), but in `o2.sh`, only the "tag" line was changed.

TimoWilken:
Hi @ktf,

> We must make sure, however, that such a tag lives on a branch (say v20.49-patches).

Could you clarify what you mean by this? Should the script use/create a branch in alidist like vYY.WW to push the tag to?

TimoWilken:
@ktf The latest commit also pushes the auto-generated commit to the requested `ALIDIST_BRANCH`. Is that what you meant?

TimoWilken:
@ktf also, if this is the intended behaviour, what should happen if the upstream `ALIDIST_BRANCH` has diverged during the build? `git push --force $ALIDIST_BRANCH` seems too destructive, unless this script is the only thing writing to that branch.

ktf:
what we usually do is to build in an "rc/something" branch. So we could push there (with force push) and have a PR opened via a github action.

TimoWilken:
@ktf, I've changed the script to push to `rc/$ALIDIST_BRANCH`. The action in alisw/alidist#2735 then takes that and creates a PR.

TimoWilken:
I've tested this PR and alisw/alidist#2735 with [a dummy "package"](https://github.com/TimoWilken/action-test). In that package repo, I set the GH action from alisw/alidist#2735 up to create PRs from `rc/` branches, which works (there's a PR open there). `daily-tags.sh` from this PR also worked in my test, using [a dummy package definition in alidist](https://github.com/TimoWilken/alidist/blob/74738612cb086eb31a46cee806b9b6289fbb3468/test.sh).

(I set the GH action up in the package repo, not in my alidist as I didn't want to make too much of a mess in my alidist fork. `daily-tags.sh` pushes an rc/ branch to both the package repo and alidist though, so it doesn't make a difference for testing the GH action.)

@ktf: Together with alisw/alidist#2735, this PR should be ready to merge in my opinion.

ktf:
Can you please squash the commits as it makes sense?

TimoWilken:
@ktf I've reorganised the commits so related changes are together.

PR #914, opened 2020-12-15
Shorter build message

ktf:
So that we actually have better chances of reading the full
host spec.

### DIFF
--- a/ci/build-loop.sh
+++ b/ci/build-loop.sh
@@ -38,7 +38,7 @@ done
 if [ "$BUILD_TYPE" = untested ]; then
   # Set a status on GitHub showing the build start time, but only if this is
   # the first build! Rebuilds should only set the final success/failure.
-  report_pr_errors --pending -m "Building since $(date +'%Y-%m-%d %H:%M %Z') on $host_id"
+  report_pr_errors --pending -m "Started on $(date +'%m/%d %H:%M') @ $host_id"
 fi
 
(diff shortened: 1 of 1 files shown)


ci/build-loop.sh
```diff
@@ -38,7 +38,7 @@ done
 if [ "$BUILD_TYPE" = untested ]; then
   # Set a status on GitHub showing the build start time, but only if this is
   # the first build! Rebuilds should only set the final success/failure.
-  report_pr_errors --pending -m "Building since $(date +'%Y-%m-%d %H:%M %Z') on $host_id"
+  report_pr_errors --pending -m "Started on $(date +'%m/%d %H:%M') @ $host_id"
```
TimoWilken, line 41: Looks good, though I'd keep some way of indicating the time zone -- the CI builders are all on UTC and I've had people contact me before I added it who were confused about the UTC times. A ["Z"][zulu] after the time (so `%H:%MZ`) might suffice though?

Another minor thing is that `%d/%m` would look a little less confusing to me than `%m/%d`, but I don't really mind either way.

[zulu]: https://www.timeanddate.com/time/zones/z "Zulu time"

ktf: How about:

```
Started Tue 18:30Z
```

I guess people will complain in any case if it takes more than a day...

TimoWilken: Good point! :)

PR #919, opened 2021-01-05
Deploy aliBuild 1.7.1

TimoWilken:


ktf:
Did you verify it actually works fine?

pzhristov:
It works fine, but there are some dangling links on Ubuntu 18.04. They are caused by the PR https://github.com/alisw/alibuild/pull/638 and I am not sure this is intentional.

TimoWilken:
Works fine on Centos 8. The dangling symlinks were supposed to fix a separate issue. Are they causing problems?

pzhristov:
@TimoWilken No, they do not create problems, but pollute the directory where I call aliBuild from.

TimoWilken:
@pzhristov Oh right, I'll have a look if I can stop it doing that! It seems to be doing the same thing on the Centos builders, but it doesn't really matter there.

TimoWilken:
@pzhristov alisw/alibuild#641 should fix the symlink problem!

TimoWilken:
I've tested v1.7.1 by testing the changes for alisw/alibuild#641. OK to deploy, @ktf?

pzhristov:
> @pzhristov [alisw/alibuild#641](https://github.com/alisw/alibuild/pull/641) should fix the symlink problem!

Thanks a lot, the issue with the simlinks is solved (tested on Ubuntu 18.04).

PR #920, opened 2021-01-07
Silence false positive linker warnings

TimoWilken:
Don't include "Warning: Unused direct dependencies" in the HTML output log. These are apparently false positives and can't be turned off at the build level.

### DIFF
--- a/report-pr-errors
+++ b/report-pr-errors
@@ -188,7 +188,10 @@ class Logs(object):
             # positives, so o2checkcode_messages is better.
             ignore_log_files=['o2checkcode-latest'])
         self.warnings_log = chunk_command_output_by_log(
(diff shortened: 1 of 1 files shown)


ktf:
Are they really false positives?

davidrohr:
The come from the CMake LINK_WHAT_YOU_USE flag, if some dependencies are not required. But I think it will be impossible to clean up the dependencies, since they come via the target interfaces, and sometimes one does not need all of the libraries that come with the target. I don't see a way to solve that. And I didn't find a way to suppress that linker warning selectively, I could only suppress all of them. So I think filtering it at the log level is the best approach in this case.

davidrohr (approved):

PR #921, opened 2021-01-07
Sync packages from S3 in aliPublish

TimoWilken:
This patch changes aliPublish to get packages from S3 instead of ali-ci.cern.ch.

To keep the package processing code intact and reduce the potential for bugs, essentially insert a shim between S3 and code that expects ali-ci's JSON format for directory listings.

Untested for now.

### DIFF
--- a/publish/aliPublish
+++ b/publish/aliPublish
@@ -481,7 +481,7 @@ def sync(pub, architectures, endpointUrl, bucket, baseUrl, basePrefix, rules,
 
   # Template URLs
   packNamesPathTpl = "%(arch)s/dist-direct"
-  distPathTpl      = "%(arch)s/dist%(dist)s/%(pack)s/%(pack)s-%(ver)s"
+  distPathTpl      = "%(arch)s/%(dist)s/%(pack)s/%(pack)s-%(ver)s"
   verPathTpl       = "%(arch)s/dist-direct/%(pack)s"
   getPackUrlTpl    = ("%(baseUrl)s/%(basePrefix)s/%(arch)s/dist-direct/%(pack)s"
                       "/%(pack)s-%(ver)s/%(pack)s-%(ver)s.%(arch)s.tar.gz")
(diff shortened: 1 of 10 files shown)


TimoWilken:
The workflow run failed, but #922 should fix that. I've run the "publish" step of the workflow successfully on my machine.

TimoWilken:
@ktf I've tested these changes with `python2 aliPublish -n sync-cvmfs -c aliPublish.conf --debug`, and it reads the package lists fine.

The only problem is that I can't see a way to get the file modification time from S3, so aliPublish just wants to publish everything it finds. (I faked an mtime as whenever the script runs.)

TimoWilken:
I've pushed a commit that gets packages' mod times from S3 using s3cmd, but I'm not sure it's perfectly reliable -- e.g. lots of Alien-CAs packages, back to v1-1, have mtimes on 05 Jan 2021, which was presumably when they were last touched on S3. (Using s3cmd on each package is also pretty slow.)

If mtimes are actually needed, then I think the best option would be to generate directory indices and upload them to S3.

@ktf: how badly does the script need mtimes? Is it just a failsafe to make sure it doesn't publish too many packages?

ktf:
@TimoWilken yes, it's a failsafe. We could think of a better way of doing it, e.g. giving a rule which matches only the last 2 months worth of releases.

TimoWilken:
@ktf: I don't like the idea of changing the matching rules to exclude old releases, because it's a bit fragile to do with regexes, and we'd presumably have to update the list every month to remove older releases.

I also ran into another problem: the HTTPS API only returns a maximum of 9999 results, but some directories contain more entries (recursively) than that, e.g.
```bash
$ curl -s 'https://s3.cern.ch/swift/v1/alibuild-repo/?prefix=TARS/slc7_x86-64/dist-direct/' | wc -l
9999
```
I've tried adding `&limit=999999` to the URL, but that just caps it even lower (at 985, for whatever reason). Is there an easy way to remove the limit? If not, I can look into using boto3 or maybe s3cmd to get the file lists instead.

ktf:
Can we start the process of making this non-draft? Can you start opening separate PRs with bits which can enter already without disrupting production?

TimoWilken:
Superseded by #967 and #972.

PR #923, opened 2021-01-08
Reduce false positives in log output

TimoWilken:
This replaces grepping for `***` with a more precise search for make/gmake errors (e.g. when a compiler is killed), a type of cmake error, and failed tests, which is what that pattern is for. This means disabled tests don't show up in the HTML error log any more.


davidrohr (approved):
Thx, looks correct for me, but I didn't test the expression on some real output :)

TimoWilken:
I've tested it on all the logs lying around on the CI servers. It catches compiler, linker, make/gmake and cmake errors and failed tests fine. :)

PR #924, opened 2021-01-11
Ignore CMAKE_CXX_COMPILER warning in HTML log

TimoWilken:
Should be disabled anyway, so remove the clause that looks for it.

Previously, we found the following messages by looking for lines of 79 asterisks:

```
*******************************************************************************
*----------------------------------- ERROR -----------------------------------*
* The variable 'CMAKE_CXX_COMPILER' should only be set by the cmake toolchain,
* either by calling 'cmake -DCMAKE_CXX_COMPILER="/opt/rocm/hip/bin/hipcc"' or
* set in a toolchain file and added with 
* 'cmake -DCMAKE_TOOLCHAIN_FILE=<toolchain-file>'.
*-----------------------------------------------------------------------------*
*******************************************************************************
```

### DIFF
--- a/report-pr-errors
+++ b/report-pr-errors
@@ -183,7 +183,7 @@ class Logs(object):
 
         self.errors_log = chunk_command_output_by_log(
             r"grep -EC 3 ': (internal compiler |fatal )?error:|^Error(:| in )|"
-            r"make.*: \*\*\*|\*\*\*Failed|^\*{79}$' -- %s",
+            r"make.*: \*\*\*|\*\*\*Failed' -- %s",
             # Errors from this log are reported in o2checkcode_messages
             # already, so don't report them twice. This log also contains false
             # positives, so o2checkcode_messages is better.


ktf:
Shouldn't we fix the error in this particular case? It's not a warning. What is actually happening?

davidrohr:
@ktf : It is not an error. AMD complains about something that is totally valid in CMake. There is an option to disable the error in the AMD CMake, but this just changes it from a CMake fatal to CMake message, so it is still printed out. So it is just a message text string that says error, which we were grepping for. I have reported the false behavior to AMD, but will take a while until it is fixed in a release. On top of that, this check should not be executed on our CMake code anyway, but only within the AMD CMake internally. So this is a second bug in the AMD CMake that the check triggers at all in our code, which I have also reported to them, but again will take a while to be fixed.
Bottom line, let's just not grep for it :)

PR #925, opened 2021-01-12
Minor improvements to list-branch-pr

TimoWilken:
# Build older PRs first

For new (yellow) PRs, older ones have been in the queue longer and should thus be built first in the interest of fairness.

For previously tested (red and green) PRs, we take a single random PR to test, so the sorting has no effect.

# Only fall through to green PRs if there are any

This modifies the check slightly so that if there are no green or yellow PRs, a red one is always built. Previously, there was a 30% chance of falling through the check for red PRs, even if there were no green ones, leading to a 10-minute sleep.

### DIFF
--- a/list-branch-pr
+++ b/list-branch-pr
@@ -277,7 +277,10 @@ def main(args):
         prs = grouped[group]
         if number is not None and number <= len(prs):
             prs = random.sample(prs, number)
-        for pull in prs:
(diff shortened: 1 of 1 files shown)


sawenzel (approved):
This sounds good to me.

davidrohr (commented):
just for reference, thx, makes totally sense

PR #928, opened 2021-01-18
Add missing system requirements for pip dependencies

TimoWilken:
python-ldap requires libldap2-dev; using other packages' pre-built binaries requires python3-wheel.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -16,7 +16,10 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Install prerequisites
-        run: pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


ktf:
Can we disable whatever requires python-ldap?

TimoWilken:
It's only used by `ci/sync-egroups.py`, which only seems to be called from `ci/run-http.sh`. I can't find an aurora definition for `run-http.sh`, but it's currently running and creating commits in ali-marathon. The last time `groups.yml` was updated was on 23 Jun 2020.

<https://github.com/alisw/ali-bot/blob/6e645b8ede148d681b51c2da8bf00898a97a1b29/ci/run-http.sh#L62>

It seems still in use to me.

PR #930, opened 2021-01-19
Use custom warning extraction code, replacing grep -P

TimoWilken:
This is because `grep -P` is unavailable on MacOS.

I've replaced the grep call with an awk script that greps for the lines we want, excludes unwanted lines, and then calculates which lines to include for context.

I've tested the awk script on Linux and using `/usr/bin/awk` on the MacOS builders.

### DIFF
--- a/report-pr-errors
+++ b/report-pr-errors
@@ -189,10 +189,20 @@ class Logs(object):
             # positives, so o2checkcode_messages is better.
             ignore_log_files=['o2checkcode-latest'])
         self.warnings_log = chunk_command_output_by_log(
-            # We need -P for (?!...) negative lookaheads, which we use to
-            # filter out a specific linker error that can't be turned off.
-            # Other warnings should go through, though.
-            "grep -PC 3 ': warning:|^Warning: (?!Unused direct dependencies:$)' -- %s",
+            # We need this complexity to filter out a specific linker warning
+            # that can't be turned off. This would be much easier with grep -P,
+            # but that doesn't work on MacOS.
+            # grep prints line numbers delimited by ":", so set -F: in awk.
+            "< /dev/null awk -F: -v logfile=%s 'BEGIN {" + """
+            # This command extracts wanted lines from the log, with their line numbers.
+            grepcmd = "grep -Ehn ': warning:|^Warning: ' -- " logfile " | " \\
+              "grep -Ev '^[0-9]+:Warning: Unused direct dependencies:$'"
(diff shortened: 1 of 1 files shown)


report-pr-errors
```diff
@@ -189,10 +189,20 @@ class Logs(object):
...
             ignore_log_files=['o2checkcode-latest'])
         self.warnings_log = chunk_command_output_by_log(
-            # We need -P for (?!...) negative lookaheads, which we use to
-            # filter out a specific linker error that can't be turned off.
-            # Other warnings should go through, though.
-            "grep -PC 3 ': warning:|^Warning: (?!Unused direct dependencies:$)' -- %s",
+            # We need this complexity to filter out a specific linker warning
+            # that can't be turned off. This would be much easier with grep -P,
+            # but that doesn't work on MacOS.
+            # grep prints line numbers delimited by ":", so set -F: in awk.
```
ktf, line 195: No way... ;-)

ktf: There must be a better way?

TimoWilken: I hope so, but I haven't found it yet...

The problem is that we want context lines, so a simple `grep | grep -v` doesn't work at all (it would keep the context lines for the unwanted warnings, which is even more confusing than just keeping the unwanted warnings).

Maybe there's a way to simulate a negative lookahead with extended regexes, but it seems even more gotcha-prone than this... e.g. `^Warning: ([^U]([^n](and so on)|$)|$)` is wrong for a few reasons.

The other options I can think of are to:
1. just include the unwanted warnings in the HTML log
2. ignore all the `Warning: ...` lines, not just `Warning: Unused direct dependencies:`
3. work out all possible `Warning: ...` messages that we *do* want to include and grep for them, instead of trying to exclude the one we don't want.

ktf: Can't we just use `perl` directly?

ktf: https://apple.stackexchange.com/questions/296834/perl-regex-in-mac-terminal-grep-p-t-t2mypattern-t-textfile-txt-doe

TimoWilken: Possibly, though the problem is, again, the context lines. I don't think it'd be a perl one-liner, as I'd have to keep track of the last 3 lines and print them all when a match is found, and then also handle the case where two matching lines are so close together that their context lines overlap. This would probably be at least as complicated as the awk script, which handles these cases fine.

TimoWilken:
Closing in favour of #965.

PR #931, opened 2021-01-20
Test QC-dataflow against O2 dev branch

TimoWilken:
This brings build/QualityControl/o2-dataflow in line with build/QualityControl/o2, which is already tested against O2@dev.


TimoWilken:
@davidrohr @Barthelemy I don't know if you're the right people to ask about this, but what about Configuration, InfoLogger, Monitoring and ReadoutCard? Should they be tested against O2@dev too?

davidrohr:
Could you do the same for the o2-dataflow test of QC on MacOS?

TimoWilken:
@davidrohr done, thanks for the reminder!

davidrohr:
> @davidrohr @Barthelemy I don't know if you're the right people to ask about this, but what about Configuration, InfoLogger, Monitoring and ReadoutCard? Should they be tested against O2@dev too?

Personally I would say yes. @ktf @sawenzel  : What do you think? Perhaps we should discuss this in WP3?

Barthelemy:
I agree, it should be raised in WP3. @ihrivnac

Barthelemy:
@TimoWilken Thank you very much for taking care of the QC tests

ihrivnac:
> I agree, it should be raised in WP3. @ihrivnac

The agenda for the next WP3 meeting is now available at: https://indico.cern.ch/event/999182/
Please, add a discussion item in the document.

TimoWilken:
Thanks @ihrivnac, I've added a discussion item.

davidrohr:
@TimoWilken : Why does the CI fail here? I'd like to merge this to fix the failures in the QC CI.

TimoWilken:
@davidrohr it needs some packages installed, #932 should fix it. When that's merged, I can rebase on top of that.

The CI doesn't check the .env files at all, though, so it shouldn't affect it in this case.

TimoWilken:
@davidrohr rebased, checks are green now.

PR #936, opened 2021-01-25
Verify interference of updateable RPMS wrt non-updateable ones

ktf:
Sometimes files which are removed in the updateable RPMs are also removed in the non-updateable ones. If this is the case, we should at least have them present.

### DIFF
--- a/publish/pub-rpms-template.sh
+++ b/publish/pub-rpms-template.sh
@@ -122,10 +122,10 @@ pushd unpack_rpm
   fi
   # Remove useless files conflicting between packages
   if [ ! "X$RPM_IS_UPDATABLE" = X ]; then
-    rm -rfv $RPM_ROOT/.build-hash            \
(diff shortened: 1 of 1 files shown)


ktf:
@TimoWilken could you check and deploy this (and get rid of the EOS sync for the RPMs?).

TimoWilken:
@ktf is a `.%(package)s` enough to stop conflicts between RPMs? Using `.%(package)s.%(version)s` or something like that seems safer. Or is it impossible to install multiple versions of the same non-updatable package side-by-side?

It also seems a bit cleaner to me to have these files in `/usr/share/%(package)s/` or similar, instead of just the root, but that's just bikeshedding. :)

ktf:
%(version)s is not needed, because those RPMs are meant to update from the previous version and the non-updateable ones have a root of the package which is not common.

For the different path, there might be mechanisms in alibuild which rely on those files being in the root installation path. They are in any case in /opt/alisw, not in /.

PR #937, opened 2021-01-25
Inline cvmfs/test-alienv.sh into workflow

TimoWilken:
This solves a security issue where a PR could change the file, and it would get executed by the PR checker with secrets in the environment.

Also use `pull_request` trigger instead of `pull_request_target`.

@ktf, is `cvmfs/test-alienv.sh` used anywhere else? Should it be kept separately?


ktf:
I do not think so. We should probably understand how much of the test is still relevant.

PR #939, opened 2021-01-28
Escape version and tag when editing them

ktf:


ktf:
@TimoWilken Don't we need this?

ktf:
Yes, we do.

PR #944, opened 2021-02-08
Add missing quotes

ktf:


ktf:
@awegrzyn #903 broke the RPMs due to the missing quotes.

awegrzyn:
Ok thanks, installing `alisw-aliswmod-4` ends up in:
```
Transaction check error:
  file /bin from install of alisw-aliswmod-4-1.el7.x86_64 conflicts with file from package filesystem-3.2-25.el7.x86
```

awegrzyn:
Apparently it tries to create `/bin` dir
```
# rpm -qpl alisw-aliswmod-4-1.el7.x86_64.rpm 
/bin
/etc/profile.d/99-aliswmod.sh
```

PR #946, opened 2021-02-09
Publish RPMs on CentOS 8

TimoWilken:
Publish the same things on CC8 as on CC7.

### DIFF
--- a/publish/aliPublish-rpms.conf
+++ b/publish/aliPublish-rpms.conf
@@ -74,6 +74,73 @@ architectures:
         - -38c51d6edf-
         - -8112bb1d4c-
 
+  slc8_x86-64:
+    RPM: el8.x86_64
+    include:
+      Toolchain:
+        - ^v.*$
+      rpm-test: True
+      mesos-workqueue: True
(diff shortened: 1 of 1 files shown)


davidrohr:
@TimoWilken : How are the CC8 RPMs built? Inside the slc8-builder, or the slc8-gpu-builder?
We need the latter, to have the GPU part built for the EPNs.

TimoWilken:
@davidrohr currently they're built on plain slc8. Thanks for the heads-up, I'll talk to Giulio tomorrow to work out a good way to build on slc8-gpu instead!

ktf:
I would just use the slc8-gpu-builder image. What is stopping us from doing so?

TimoWilken:
Is it OK if all the slc8 Jenkins builds run on slc8-gpu? If so, that makes things a bit simpler.

ktf:
yes. That's fine.

TimoWilken:
Great, done!

TimoWilken:
Thanks! I'll restart the rpm_creation task so it picks up these changes.

PR #947, opened 2021-02-09
Revert dropping aliswmod #903 #944

awegrzyn:
Unfortunately, the old RPM pickup new `aliswmod`, I thought all deps are fixed, but `aliswmod` is the exception. Let's revert it for the moment and find out how to do it in a clean way.

### DIFF
--- a/publish/pub-rpms-template.sh
+++ b/publish/pub-rpms-template.sh
@@ -25,26 +25,42 @@ if [[ $RPM_IS_UPDATABLE ]]; then
 fi
 
 # Create aliswmod RPM
-ALISWMOD_VERSION=4
+ALISWMOD_VERSION=3
 ALISWMOD_RPM="alisw-aliswmod-$ALISWMOD_VERSION-1.%(arch)s.rpm"
 if [[ ! -e "%(repodir)s/$ALISWMOD_RPM" ]]; then
   mkdir -p aliswmod/bin
   mkdir -p aliswmod/etc/profile.d
-  cat > aliswmod/etc/profile.d/99-aliswmod.sh << EOF
-export LD_LIBRARY_PATH=/opt/alisw/el7/lib:/opt/alisw/el7/lib64:\$LD_LIBRARY_PATH
-export PATH=/opt/alisw/el7/bin:\$PATH
+  cat > aliswmod/etc/profile.d/99-aliswmod.sh << \EOF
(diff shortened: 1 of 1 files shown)


awegrzyn:
@ktf Could you also please remove `RPMS/el7.x86_64/alisw-aliswmod-4-1.el7.x86_64.rpm` from S3?

awegrzyn:
@TimoWilken could you help with merging this and making sure RPM builder pick this up?

TimoWilken:
OK, merged. I've restarted the RPM builder as well.

awegrzyn:
Reason for this revert
- As `aliswmod` version is not fixed on RPM level, old RPMs pick up new version as well
- `aliswmod` v4 tried to create `/bin` dir what resulted in transaction error

Instead that we should
- Drop entirely `aliswmod` for new "non-updatable" RPMs

ktf:
Maybe we should simply rename the rpm to something else? I guess we do need the environment no?

awegrzyn:
The environment is not needed for "non-updatable" RPMs.
For updatable RPMs we don't really want to for the moment as our packages break system tools, so it's better to set paths only once this is needed (eg. in ControlWorkflows).

ktf:
How about we clean completely the template from the updatable RPMs and we have a separate template, just for those? I fear two many issues are related to trying to kill two pigeons with one stone.

awegrzyn:
sure, I think it's very good idea

awegrzyn:
@TimoWilken are you sure RPM builder uses new commit? An RPM that has been just created uses `RPMS/el7.x86_64/alisw-aliswmod-4-1.el7.x86_64.rpm`.

TimoWilken:
The rpm_creation task has ali-bot 30a2c29, so it's got the changes from this PR. Could that be a leftover from before I restarted it?

awegrzyn:
ok thanks, idk, but we just need to make sure that `aliswmod-4` does not get produced.

vascobarroso:
@TimoWilken @ktf please treat this as high priority, it is currently blocking us from progressing with the preparations for MW2.

ktf:
`aliswmod-4` is not being produced. I guess this is a leftover from the previous runs? The new one is called `aliswmod4`

ktf:
@vascobarroso it is top priority.

vascobarroso:
@ktf great, thanks a lot for the support 👍

ktf:
I love the smell of code when it gets deleted.

PR #948, opened 2021-02-09
Disable ALIBUILD_O2_TESTS in fullCI

TimoWilken:
This is so that the fullCI build gets past the currently failing tests and can check other things.

This also disables `-Werror` in fullCI, but other O2 builds still check whether it compiles without warnings.


davidrohr:
Merging

- There is no reason to run the tests twice, the normal CI will run the same tests. And the GPU tests are anyway disabled in the full CI, since the CI machines have no GPUs.
- This should speed up the full-CI, which is the slowest, since it tests more things than the others.
- This covers also the code patchs without `-DNDEBUG`.
- This avoids that the full-CI becomes red with a randomly failing test, and then gets merged commenting that the failure is unrelated, since in that case the full-system-test and the code-checker do not run. Thus it is favorable to make the full CI continue in any case, also when a test fails, to make sure these additional tests are running.
I'd therefore propose that we leave it this way, also when the current CI problems are fixed.

PR #950, opened 2021-02-10
Add separate template for updatable RPMS

ktf:


ktf:
This simplifies the template for the RPMs, hopefully fixes the dropping of init.sh files and makes sure the aliswmod package is called aliswmod4, to indicate in requires module 4 as a dependency.

ktf:
There is a typo. Do not merge. Fixing it.

ktf:
@awegrzyn FYI.

PR #951, opened 2021-02-10
Only run non updatable RPMS

ktf:


ktf:
Do not try to run the updateable RPMS for now

PR #952, opened 2021-02-10
Avoid spending ages in deleting the staging area

ktf:


ktf:
This should also speedup considerably the RPM creation...

PR #963, opened 2021-02-15
Reduce JOBS to 6 in CI and Jenkins builds

TimoWilken:
This is to avoid the compiler segfaults.


ktf:
It says 10...

TimoWilken:
I've fixed the title!

davidrohr:
Does this really fix the compiler segfault? A segfault is usually not triggered by too little memory.

ktf:
I did not realise it was a segfault. Indeed we should try to understand what is going on...

PR #964, opened 2021-02-16
Add fullCI check for alidist

TimoWilken:
This simply copies the check in O2 to also be done in alidist, checking against O2@dev.

The new check is called `build/O2/fullCI/alidist`.

I've extracted the parts common to the slc8-gpu container into the relevant `DEFAULTS.env`.


davidrohr:
Thx, looks good to me.

PR #966, opened 2021-02-16
[O2-2045] Drop aliswmod from non-updateable RPMs

awegrzyn:


awegrzyn:
To be merged when MW2 is over.

ktf (approved):

ktf:
Shall we go ahead?

awegrzyn:
For me that's OK, although I don't know the impact on C8. It comes with `environment-modules-4.5.2-1` so in general should be ok.

PR #967, opened 2021-02-16
Update publishing script with setup and canary files

TimoWilken:
The new script also pulls changes to ali-bot on every iteration, and loops faster (sleeping only a minute between iterations).

This is to simplify the aurora service definition.

### DIFF
--- a/publish/run-all-rpms.sh
+++ b/publish/run-all-rpms.sh
@@ -1,7 +1,31 @@
-#!/bin/bash
+#!/bin/bash -ex
 
-# Check if variables are provided.
-[ X${DEBUG:-1} = X1 ] && set -x
+# The architecture should correspond to the architecture built using the
+# requested config files -- only these architectures' repos will be synced.
+arch=$1
+shift
+# $@ now contains only the config files to read.
 
-python -m cProfile -s cumtime ./aliPublish --config "aliPublish-rpms.conf" --debug --cache-deps-dir /tmp/pubdepscache sync-rpms >&2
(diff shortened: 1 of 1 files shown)


ktf:
Sorry, can we keep it easy and simply have the commands we need to run one after the other? What is the advantage of the looping?

TimoWilken:
@ktf is this better?

TimoWilken:
I'll have to change this if #972 is merged, so marking as draft for now.

publish/run-all-rpms.sh
```diff
@@ -1,7 +1,47 @@
...
+# look for the RPMs they expect.
+# We build this file list now as aliPublish can take some time, and other builds
+# (that we don't want to notify yet) might finish in the meantime. We run
+# aliPublish only for one specific architecture, so we have
+# architecture-specific canary files under rpmstatus/$arch/.
+s3cmd ls "s3://alibuild-repo/rpmstatus/$arch/" | cut -b 32- > canaries.txt
+
+case "$arch" in
+  el8.*)
+    pip install boto3    # aliPublishS3 needs boto3.
```
ktf, line 31: Can you put the dependency in `setup.py`?

ktf (approved):

ktf:
We deploy tomorrow.

publish/run-all-rpms.sh
TimoWilken: It's already there. However, `setup.py` (and `requirements.txt`) also has stuff we don't need here, like `python-ldap`, `PyGithub` and `klein`. If you prefer, I can put a `pip install -r ../requirements.txt` in the script instead, though.

TimoWilken: (We don't install ali-bot here, we just clone it, as we need the config files from it, and if we're cloning it already, we might as well just use the scripts from the repo.)

PR #970, opened 2021-02-16
Use unique staging directory

TimoWilken:
To avoid interference between multiple aliPublish instances.


TimoWilken:
Not really needed, I suppose. CC7 instances can keep using the original directory; CC8 ones will use unique ones.

PR #971, opened 2021-02-16
Publish RPMs from S3

TimoWilken:
- use boto3 to fetch packages on S3 (HTTPS API doesn't return the whole list)
- file modification times are now ignored -- this was a failsafe before
- S3's "fake" symlinks are handled properly
- aliPublish*.conf are updated to match the changes to aliPublish itself
- a minimum of unused code is removed -- better cleanup in separate PR

@ktf: let me know if this diff is still too big to review. In that case, I'll try to split it up a bit more (though that's difficult because most of the changes in this PR are fairly tightly coupled).

### DIFF
--- a/publish/aliPublish
+++ b/publish/aliPublish
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-import logging, sys, json, yaml, requests, urllib3, errno
+import logging, sys, json, yaml, requests, errno, boto3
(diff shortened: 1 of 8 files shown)


ktf:
I would simply have an `aliPublishS3` and use that to publish cc8, with no other changes to CC7 which we cannot afford to break right now.

PR #973, opened 2021-02-19
Extract full-system-test-specific log messages

TimoWilken:
Adds a "O2 full system test" section to the HTML log, which shows up if either of these types of problem was detected:
- `task timeout reached`
- `Detected critical problem in logfile`


davidrohr (approved):
Looks good, let's try it

PR #974, opened 2021-02-19
After build, wait for RPMs to be published

TimoWilken:
This script is intended for use in Jenkins. It creates a canary file that the RPM publisher removes when done, then checks if an RPM was created as expected, tries to install it, and reports success if this check is OK.

Untested so far.

### DIFF
--- a/publish/check-for-published-rpm.sh
+++ b/publish/check-for-published-rpm.sh
@@ -6,7 +6,7 @@ arch=$1
 
 lsrpm () {
     # Show published RPMs for $PACKAGE_NAME.
-    s3cmd ls "s3://alibuild-repo/RPMS/$arch/alisw-$PACKAGE_NAME+" |
+    s3cmd ls "s3://alibuild-repo/RPMS/$arch/alisw-${PACKAGE_NAME:?}+" |
         # Trim leading modtime, file size, URL path and file extension, leaving
         # only the package name, which we can pass to yum.
         sed 's|^.*/||; s|\.rpm$||' | sort  # comm(1) expects lines in sorted order.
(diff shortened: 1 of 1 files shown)


publish/check-for-published-rpm.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Store list of RPMs for this package, so we can compare later.
+lsrpm > old-rpms.txt
+
+# Upload a canary file; this will be removed by the publishing script once it's
+# done, so we know when to check for our desired RPM. aliPublish has an instance
+# per architecture, so make per-architecture canary files.
+canary=$arch/$BUILD_TAG.finished
+date | s3cmd put - "s3://alibuild-repo/rpmstatus/$canary"
+while [ -n "$(s3cmd ls "s3://alibuild-repo/rpmstatus/$canary")" ]; do
```
ktf, line 23: Do you really need to wait so long?

publish/check-for-published-rpm.sh
```diff
@@ -0,0 +1,56 @@
...
+# Show only lines in "file 1", i.e. filenames that are only in S3.
+lsrpm | comm -23 - old-rpms.txt > new-rpms.txt
+
+# Check if there are any new RPMs.
+if [ -z "$(cat new-rpms.txt)" ]; then
+    echo "FAILED: aliPublish finished, but produced no new RPM for $PACKAGE_NAME" >&2
+    exit 1
+fi
+
+echo "SUCCESS: New RPMs for $PACKAGE_NAME follow:" >&2
```
ktf, line 37: I would drop "SUCCESS" since you do not know if the package is actually installable at this point.

publish/check-for-published-rpm.sh
TimoWilken: Done!

publish/check-for-published-rpm.sh
TimoWilken: Done!

ktf (approved):

ktf:
Looks good to me. Why is it still draft?

PR #977, opened 2021-02-24
Increase ulimit on MacOS

TimoWilken:
Git seems to need this for fetching large repos.


davidrohr:
@ktf @TimoWilken : I just saw a very similar issue with the ulimit in the full CI. The full system test needs to open plenty of files. Actually 1024 will not be enough. Can we increase the ulimit to 4096, at least for the full CI builders? (O2 and alidist)

PR #978, opened 2021-02-25
Only increase ulimit if needed

TimoWilken:
The largest allowable limit seems to be 10240 on alibuildmac00. The others have a maximum limit of 24576, and on Linux the maximum limit is even higher. This patch doesn't change the ulimit if it is already large enough, as the O2 full system test on Linux needs a very high ulimit.


ktf (approved):

ktf:
great, thanks, can you reset the fullCI test?

TimoWilken:
@ktf if it's alright, I'll restart ci_alisw_slc8-gpu-builder_latest/2 now, as it restarted yesterday and picked up the previous (lower) ulimit setting.

ktf:
yes, please.

PR #979, opened 2021-02-25
Fix handling of utf8

ktf:


ktf:
@pzhristov this should make the user syncing work again.

PR #981, opened 2021-02-26
Reduce JOBS for fullCI

TimoWilken:
It seems to be running out of memory again...

See e.g. https://github.com/AliceO2Group/AliceO2/pull/5569.


davidrohr:
merging for now, but we should really see what we can do to run with more jobs, to speed up the CI.

PR #984, opened 2021-03-02
Use yum's test mode to install generated RPMs

TimoWilken:
`yum --assumeno` exits with an error, even if the transaction would have succeeded. The test mode is more comprehensive and has the correct exit code.

### DIFF
--- a/publish/check-for-published-rpm.sh
+++ b/publish/check-for-published-rpm.sh
@@ -25,7 +25,7 @@ while [ -n "$(s3cmd ls "s3://alibuild-repo/rpmstatus/$canary")" ]; do
 done
 
 # Now see if the RPM we want has been published.
-# Show only lines in "file 1", i.e. filenames that are only in S3.
+# Show only lines in "file 1", i.e. only new filenames.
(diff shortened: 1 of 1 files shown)


publish/check-for-published-rpm.sh
```diff
@@ -48,9 +48,16 @@ EOF
...
+if [ $yumerr -gt 0 ]; then
+  echo 'FAILED: could not install new RPMs. See above for errors from yum.' >&2
+else
+  echo 'SUCCESS: no errors reported when installing new RPMs (dry-run only)' >&2
 fi
 
-echo 'SUCCESS: no errors reported when installing new RPMs (dry-run only)' >&2
+# We don't need the downloaded RPMs any more, so clear the cache to avoid
+# accumulating crud.
+yum clean packages || true
```
ktf, line 62: This will still keep them installed, though, no?

TimoWilken: They are never actually installed in the first place -- `yum install --setopt tsflags=test` makes sure they *could* be installed, downloads the packages, but doesn't actually install them.

PR #985, opened 2021-03-02
Improvements to Jenkins daily build script

TimoWilken:
- parse `AUTOTAG_PATTERN` before build starts

  This avoids bugs where waiting for PRs to be merged can make the build roll over midnight, which leads to `AUTOTAG_TAG` holding the wrong date.

- separate steps run in non-overlapping `node` blocks

  This means a slot of the slc7-light builder is not taken up by a daily build that is not waiting for pull requests any more. If waiting for PRs is disabled, the script also doesn't need an slc7-light slot at all any more.

- `daily-tags.sh` now uses `git rev-parse` instead of parsing `git ls-remote`

  This improves pre-build setup dramatically. As we clone the package's repo anyway, we may as well get the hashes directly instead of making three additional network requests.

- removed `TEST_TAG` and `JOBS` settings

  `TEST_TAG` would previously prepend a string to the tag to be generated. This can just as easily be done using `AUTOTAG_PATTERN`, so this extra option isn't needed.

  `JOBS` was previously set depending on `MESOS_QUEUE_SIZE`, but we always run on identically-sized nodes anyway, so this complexity isn't needed.

- readability improvements and simplification

  I combined the credential setup into a single step, so it's easier to read and executes faster. Alibuild is installed in the Jenkins script, not in `daily-tags.sh`, to simplify the latter; while cleanup is done in the shell script, as the required variables are available there. I also improved quoting in the shell script.

### DIFF
--- a/daily-tags.sh
+++ b/daily-tags.sh
@@ -1,109 +1,25 @@
 #!/bin/bash -ex
-set -x
-
-# Check for required variables
-ALIDIST_SLUG=${ALIDIST_SLUG:-alisw/alidist@master}
-[ ! -z "$PACKAGE_NAME" ]
-[ ! -z "$AUTOTAG_PATTERN" ]
-[ ! -z "$NODE_NAME" ]
-
-# Clean up old stuff
-rm -rf alidist/
-
-# Determine branch from slug string: group/repo@ref
-ALIDIST_BRANCH="${ALIDIST_SLUG##*@}"
-ALIDIST_REPO="${ALIDIST_SLUG%@*}"
-
-git config --global user.name 'ALICE Builder'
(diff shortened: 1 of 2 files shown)


ktf:
This is in general fine, however there are too many changes for me to be able to review. Can you please split it up?

ktf (changes requested):
Please split it up.

TimoWilken:
First split PR: #993 (for the first change).

ktf:
Is this still relevant?

TimoWilken:
Yes, I haven't split up the other changes yet.

ktf:
Is it ready for review?

PR #986, opened 2021-03-03
Allow optional letter in O2DPG version

TimoWilken:
To enable multiple releases per day, but avoiding incrementing the number (as that would be confusing if the number is the day almost everywhere else).


ktf:
@TimoWilken who requested this change?

ktf (commented):
Who is requesting for this change?

TimoWilken:
@sawenzel is.

ktf:
Fine with me, but I guess @chiarazampolli and @fprino need to agree as well.

chiarazampolli:
Nothing against it from my side!

fprino:
It is fine also with me.

PR #989, opened 2021-03-05
Don't quit publishing job in aliPublish fails

TimoWilken:


publish/run-all-rpms.sh
```diff
@@ -40,12 +40,14 @@ while true; do
   esac
 
   for conf in "$@"; do
-    "$aliPublish" --config "$conf" --debug sync-rpms >&2
+    if "$aliPublish" --config "$conf" --debug sync-rpms >&2; then
```
ktf, line 43: What if the upload fails? will it retry?

TimoWilken: No, if `rclone` fails, it'll only retry on the next iteration.

PR #991, opened 2021-03-05
Mark build unstable if no new RPM is found

TimoWilken:
This adds an explanatory message to the script that waits for an RPM to be published. If case no new RPM is found, this may not indicate a failure of aliPublish, but the previous step might just not have produced a new tarball to publish.

### DIFF
--- a/publish/check-for-published-rpm.sh
+++ b/publish/check-for-published-rpm.sh
@@ -30,8 +30,22 @@ lsrpm | comm -23 - old-rpms.txt > new-rpms.txt
 
 # Check if there are any new RPMs.
 if [ -z "$(cat new-rpms.txt)" ]; then
-  echo "FAILED: aliPublish finished, but produced no new RPM for $PACKAGE_NAME" >&2
-  exit 1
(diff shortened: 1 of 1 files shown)


publish/check-for-published-rpm.sh
```diff
@@ -30,8 +30,18 @@ lsrpm | comm -23 - old-rpms.txt > new-rpms.txt
 
 # Check if there are any new RPMs.
 if [ -z "$(cat new-rpms.txt)" ]; then
-  echo "FAILED: aliPublish finished, but produced no new RPM for $PACKAGE_NAME" >&2
-  exit 1
+  cat >&2 << EOF
+
+WARNING: aliPublish finished, but produced no new RPM for $PACKAGE_NAME
+
+This may be because aliPublish failed, or because an existing package was
```
ktf, line 37: I would mention also the fact you might have something wrong in aliPublish-rpms.conf.

TimoWilken: OK, updated!

PR #992, opened 2021-03-08
Reduce JOBS in fullCI

TimoWilken:
Builders are out of memory again.


davidrohr:
Merging, but this is insane. We need builders with more memory.

PR #993, opened 2021-03-11
Parse AUTOTAG_PATTERN before waiting for PRs

TimoWilken:
This fixes a bug where the date rolls over while waiting for PRs, leading to an incorrectly named tag.


TimoWilken:
@ktf: tested and working.

TimoWilken:
Cc: @pzhristov -- this should fix the builds that get the wrong tag if they run over midnight.

PR #994, opened 2021-03-11
Reduce Jenkins jobs to 8

TimoWilken:
Builders are runnning out of memory, again.


ktf:
Are we sure it's not the jenkins worker that is leaking?

PR #1002, opened 2021-04-12
Manually set required build vars for CC8 O2 builds

TimoWilken:
These are already set in the Docker container, but those settings apparently aren't applied in our build containers.

### DIFF
--- a/jenkins/daily-tags
+++ b/jenkins/daily-tags
@@ -122,6 +122,15 @@ node ("slc7_x86-64-light") {
                 *slc6_x86-64*)
                   # python3 is not installed on the slc6-builder. Installing it seems to break the build.
                   pip install --ignore-installed --upgrade --user git+https://github.com/$ALIBOT_SLUG ;;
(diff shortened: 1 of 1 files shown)


davidrohr (changes requested):
Hi Timo, thx for the fix. Just few points:
- I am not sure whether we should put ENABLE_UPGRADES, that is for the full CI, but not for the container.
- Why do we change ALIBUILD_O2_TESTS? I see that it kind of makes sense to disable it. But why wasn't that done before? And it is also not disabled for the other daily-tags setups. If I'd disable it, I'd do it consistently, and not only for this container.

TimoWilken:
Hi @davidrohr, thanks for the review. I'll remove `ENABLE_UPGRADES` and `ALIBUILD_O2_TESTS` -- I just copied the variables from <https://github.com/alisw/ali-bot/blob/master/ci/repo-config/mesosci/slc8-gpu/DEFAULTS.env>.

PR #1009, opened 2021-04-22
Support python3

ktf:


ktf:
@TimoWilken this is actually not so trivial... Can you check what is needed to move aliPublish to python3?

TimoWilken:
I'll have a look. I found a `dict.iteritems()` in there (L998?), which needs to become `dict.items()`. Seems like it's not using `bytes` much, so it shouldn't be too complicated, I think.

ktf:
The code itself is straightforward actually. The issue is that we need to deploy the publisher to use python3. Can you take a look into that?

PR #1013, opened 2021-04-23
Use O2 daily tags in dataflow CI check

awegrzyn:
This is to make sure that QC RPMs are built correctly.
Currently if QC uses features from dev the RPM compilation may fail as it uses tag from O2 recipe.


awegrzyn:
@Barthelemy could you confirm?

ktf:
Ping @Barthelemy

Barthelemy (approved):

Barthelemy:
everything fails ? are we waiting for something else ?

TimoWilken:
CI errors are unrelated, #1021 should fix those. The files this PR changes aren't tested by these checks anyway.

PR #1017, opened 2021-04-29
Only include warnings from logs for main package in HTML log

TimoWilken:
Except when building alidist, in which case include all logs.

This cuts down on the number of useless messages in the HTML log.

Errors and other messages are still shown from all logs in all cases.

Untested so far -- if it's OK, I can deploy it to one builder first, before merging.

### DIFF
--- a/ci/build-helpers.sh
+++ b/ci/build-helpers.sh
@@ -106,11 +106,18 @@ function reset_git_repository () {
 
 function report_pr_errors () {
   # This is a wrapper for report-pr-errors with some default switches.
+  local extra_args=()
+  # If we're not checking alidist, only include logs for the package we're
+  # actually building in the HTML log. If we are checking alidist, all logs are
+  # potentially relevant.
+  if [ "$PR_REPO" != alisw/alidist ] && [ -n "$PACKAGE" ]; then
+    extra_args+=(--main-package "$PACKAGE")
+  fi
   short_timeout report-pr-errors ${SILENT:+--dry-run} --default "$BUILD_SUFFIX" \
                 --pr "$PR_REPO#$PR_NUMBER@$PR_HASH" -s "$CHECK_NAME"            \
                 --logs-dest s3://alice-build-logs.s3.cern.ch                    \
                 --log-url https://ali-ci.cern.ch/alice-build-logs/              \
-                "$@"
+                "${extra_args[@]}" "$@"
 }
 
 function source_env_files () {
--- a/ci/build-helpers.sh
+++ b/ci/build-helpers.sh
@@ -106,12 +106,18 @@ function reset_git_repository () {
 
 function report_pr_errors () {
   # This is a wrapper for report-pr-errors with some default switches.
-  local extra_args=()
+  local repo checkout_name extra_args=()
   # If we're not checking alidist, only include logs for the package we're
   # actually building in the HTML log. If we are checking alidist, all logs are
   # potentially relevant.
   if [ "$PR_REPO" != alisw/alidist ] && [ -n "$PACKAGE" ]; then
     extra_args+=(--main-package "$PACKAGE")
+    # In $DEVEL_PKGS, the checkout name (third field) must be a valid package
+    # name (otherwise aliBuild wouldn't recognise it as a development package).
+    while read -r repo _ checkout_name; do
+      [ "$repo" = alisw/alidist ] && continue
+      extra_args+=(--main-package "${checkout_name:-$(basename "$repo")}")
+    done <<< "$DEVEL_PKGS"
   fi
   short_timeout report-pr-errors ${SILENT:+--dry-run} --default "$BUILD_SUFFIX" \
                 --pr "$PR_REPO#$PR_NUMBER@$PR_HASH" -s "$CHECK_NAME"            \
--- a/report-pr-errors
+++ b/report-pr-errors
@@ -270,7 +270,7 @@ class Logs(object):
         self.full_system_test = self.grep_logs(
             r'Detected critical problem in logfile',
             context_before=0, context_after=20)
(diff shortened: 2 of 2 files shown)


ktf (dismissed):

TimoWilken:
Working. One question, though: e.g. for the fullCI check, we're not building O2, but O2FullCI, which doesn't do much except depend on O2. In that case, if there are no logs for `$PACKAGE`, should we show all warnings, or is it fine with no warnings at all? (Errors are still shown anyway. [Example log here](https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/AliceO2/6001/8fc0edbd40d2f249ef3b663807133a9c88b076ba/build_O2_fullCI/pretty.html>))

davidrohr:
Personally, I would always show all warnings of the package the CI is for. I.e. for AliceO2 the CI should always show all warnings compiling O2 (even though the aliBuild command is for O2FullCI), the QualityControl CI should show all warnings for QualityControl, even though building O2Suite, etc.

I don't see that showing additional warnings is needed. In any case, a warning should not break the build.

In addition, as mentioned some time ago, for the case where the build does not even arrive at the relevant package, I would like an indication which package already broke the build in the HTML log, as requested here: https://alice.its.cern.ch/jira/browse/O2-2207

TimoWilken:
> Personally, I would always show all warnings of the package the CI is for. I.e. for AliceO2 the CI should always show all warnings compiling O2 (even though the aliBuild command is for O2FullCI), the QualityControl CI should show all warnings for QualityControl, even though building O2Suite, etc.

That's a little more complicated, as we'd also have to work backwards from `PR_REPO` to find which package is the "main" one. Not impossible, but a little annoying and we'd have to parse the YAML in alidist as well.

> I don't see that showing additional warnings is needed. In any case, a warning should not break the build.

Warnings don't break the build (except when they become errors, but then they're always shown anyway).

> In addition, as mentioned some time ago, for the case where the build does not even arrive at the relevant package, I would like an indication which package already broke the build in the HTML log, as requested here: https://alice.its.cern.ch/jira/browse/O2-2207

Currently, alibuild doesn't communicate with `report-pr-errors`. We'd need e.g. a `sw/last-build-info.json`, written by alibuild, which `report-pr-errors` can read for this sort of info. Would that be alright, @ktf?

davidrohr:
> > Personally, I would always show all warnings of the package the CI is for. I.e. for AliceO2 the CI should always show all warnings compiling O2 (even though the aliBuild command is for O2FullCI), the QualityControl CI should show all warnings for QualityControl, even though building O2Suite, etc.
> 
> That's a little more complicated, as we'd also have to work backwards from `PR_REPO` to find which package is the "main" one. Not impossible, but a little annoying and we'd have to parse the YAML in alidist as well.
> 
Well, but I think this is actually the essential part: we should show the warnings for the package that is actually built.
Doesn't ali-bot know which package this is?

> > In addition, as mentioned some time ago, for the case where the build does not even arrive at the relevant package, I would like an indication which package already broke the build in the HTML log, as requested here: https://alice.its.cern.ch/jira/browse/O2-2207
> 
> Currently, alibuild doesn't communicate with `report-pr-errors`. We'd need e.g. a `sw/last-build-info.json`, written by alibuild, which `report-pr-errors` can read for this sort of info. Would that be alright, @ktf?
Aren't we already sorting the log files according to the timestamp of the file, to show them in the right order in the full log. In case the built failed, don't you know the failing package simply from the latest log?
Well, and then there is the chance that it failed not during building the package but at git checkout or so, IIRC. But in that case it would anyway be good to have a separate indication that some kind of internal error happened.

ktf:
I guess we could show all the warnings of the devel packages, apart from alidist.

davidrohr:
> I guess we could show all the warnings of the devel packages, apart from alidist.

I think that is actually the best approach!

TimoWilken:
> Well, but I think this is actually the essential part: we should show the
> warnings for the package that is actually built. Doesn't ali-bot know which
> package this is?

Well, sort of. ali-bot knows which package it should build (`$PACKAGE`), but 
this is O2FullCI, not O2, for the fullCI builds.

It also knows which repository to send statuses/comments to (`$PR_REPO`), but 
we'd have to parse alidist to find the correct package name for the repo. (Logs 
are named by package name, not by git repo name.)

> Aren't we already sorting the log files according to the timestamp of the
> file, to show them in the right order in the full log. In case the built
> failed, don't you know the failing package simply from the latest log?
> Well, and then there is the chance that it failed not during building the
> package but at git checkout or so, IIRC. But in that case it would anyway
> be good to have a separate indication that some kind of internal error
> happened.

Well, yes, if something happens before we get to the actual build (e.g. a git 
error, as you say), we don't get a log for that package at all. We'd still 
have to signal that some sort of internal error happened, and which package it 
happened in.

> I guess we could show all the warnings of the devel packages, apart from
> alidist.

Is `$PR_REPO_CHECKOUT` guaranteed to be a package name? It seems so, as O2 is 
cloned into O2, not AliceO2, and the other instances look fine. If so, this 
would make things easy.

ktf:
Is this ready to go?

davidrohr:
Just so I understand correctly, this is only for the compiler warnings, not for the other stuff we are searching for in the logs?

ktf:
@TimoWilken this now needs a rebase.

TimoWilken:
As discussed, I've changed the scripts such that warnings from all development packages are shown in the HTML log, instead of just warnings from the "main" package.

As an example, for the `build/O2/fullCI`, the new behaviour would be to show warnings only from the `O2` and `O2FullCI` builds in the HTML log.

This PR only affects which warning messages are shown in the HTML log. Other messages and the plain-text log are unaffected.

As yet untested.

PR #1019, opened 2021-05-04
Publish Focal

ktf:


ktf:
@iarsene @pzhristov: this is to publish Focal on CVMFS.

PR #1020, opened 2021-05-04
Use Python 3 for aliBuild in Jenkins and CI

TimoWilken:
This changes the CI builders and the build-any-ib and daily-tags Jenkins jobs to install alibuild with pip3 instead of pip2.

There are probably more Jenkins jobs that have to be changed manually. Is there anything else that uses alibuild that I've missed?


ktf:
@TimoWilken conflicts.

PR #1023, opened 2021-05-18
Fix CI skipping PRs in the queue

TimoWilken:
I suspect that something inside build-helpers is reading its stdin. This causes the remaining contents of `$HASHES` to be discarded, so the outer loop doesn't process the last PRs in its queue.

I also patched the other places where we use `echo | while read`, just in case.


ktf (approved):

TimoWilken:
Tested and working; merged.

PR #1024, opened 2021-05-20
Use CERN-local time zone for status message

TimoWilken:
This is less confusing than UTC.

### DIFF
--- a/ci/build-loop.sh
+++ b/ci/build-loop.sh
@@ -38,7 +38,7 @@ done
 if [ "$BUILD_TYPE" = untested ]; then
   # Set a status on GitHub showing the build start time, but only if this is
   # the first build! Rebuilds should only set the final success/failure.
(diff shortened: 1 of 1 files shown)


ci/build-loop.sh
```diff
@@ -38,7 +38,7 @@ done
 if [ "$BUILD_TYPE" = untested ]; then
   # Set a status on GitHub showing the build start time, but only if this is
   # the first build! Rebuilds should only set the final success/failure.
-  report_pr_errors --pending -m "Started $(date -u +'%a %H:%MZ') @ $host_id"
+  report_pr_errors --pending -m "Started $(TZ=Europe/Zurich date +'%a %H:%M') @ $host_id"
```
ktf, line 41:
```suggestion
  report_pr_errors --pending -m "Started $(TZ=Europe/Zurich date +'%a %H:%M CET') @ $host_id"
```

PR #1025, opened 2021-05-21
Run QC tests also on CC8

ktf:


ktf:
@TimoWilken we need to run a few FLP tests also for CC8. Can you cross check this is what is needed? We also need to deploy a separate builder, since we do not have one for the mesosdaq quota, correct? Can you do that?

ktf:
@Barthelemy

TimoWilken:
I can create a new builder (there's no slc8 one under mesosdaq yet). Any preference for the slc8 or slc8-gpu container?

The only thing I'm missing in the `.env` file would be a `JOBS=` line. I don't think there's anything for it to fall back to under `mesosdaq/slc8`.

Barthelemy:
Thank you, I don't have a preference. Basically we just want to be able to assess that the code builds properly in C8[S] the same way it does on CC7.

TimoWilken:
OK, thanks. I've started a builder under mesosdaq using the plain slc8 container, assuming `JOBS=8`.

ktf:
I've added the JOBS=8 bit. Can you cross check and merge / deploy this? @Barthelemy if you then confirm we can add the other builders.

TimoWilken (approved):

PR #1028, opened 2021-06-09
Update DEFAULTS.env

ktf:


ktf:
@TimoWilken @Barthelemy @awegrzyn I am increasing the number of CPU used for the FLP jobs to 12 (from 8). This should speed up a bit the builds. Please let's keep an eye if we get the out of memory messages again (in principle we shouldn't in any case, since you are not compiling the analysis bits).

Barthelemy:
ok, thank you, let's see

PR #1031, opened 2021-06-23
Add alidist o2-dataflow check on CentOS Stream 8

TimoWilken:
This is the only check to be newly added (instead of migrated from slc8), so use this for testing.


ktf:
do we have enough resources?

TimoWilken:
Yes, there's enough for one instance.

PR #1033, opened 2021-06-25
Increase number of jobs on slc8-gpu

TimoWilken:


davidrohr:
Are we sure that the memory issues have been solved meanwhile? Or do we hope this is solved by the new container setup? Would be pretty good if it was so.

ktf:
Well, it's a bit we do not see the, no? we can always immediately resolve. The new container setup will mitigate the issue with the hanging simulation zombies, which I think was also part of the equation.

PR #1034, opened 2021-06-28
Publish one-off releases for AliRoot

TimoWilken:
Needed for https://alice.its.cern.ch/jira/browse/ALIROOT-8707.

@ktf, is this the correct file for lxplus/Grid publishing or do I need to add it elsewhere?

### DIFF
--- a/publish/aliPublish.conf
+++ b/publish/aliPublish.conf
@@ -68,6 +68,8 @@ architectures:
       AliRoot:
        - ^v5-0[45]-Rev-.+$
        - ^v5-.*-itsmisalign-[0-9]+$
+       # One-off releases for https://alice.its.cern.ch/jira/browse/ALIROOT-8707
+       - ^v5-10-01-test-hepmc-[0-9]+$
       AliGenerators:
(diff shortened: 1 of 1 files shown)


ktf:
I wouldnt call it like that. Where was this discussed?

TimoWilken:
Nowhere, I just made that up. Any preference for how it should be called?

TimoWilken:
Updated with the new naming convention.

I've published new tags in [AliRoot] and [alidist], and removed the old `JIRA-*` tags (including in AliPhysics).

[AliRoot]: https://github.com/alisw/AliRoot/releases/tag/v5-10-01-test-hepmc-1
[alidist]: https://github.com/alisw/alidist/releases/tag/AliRoot-v5-10-01-test-hepmc-1

ktf:
Mmm, no we should keep the 09 part, not move it to 10.

TimoWilken:
What minor number are we on, then? `v5-09-01-test-hepmc-1`? `v5-09-58-test-hepmc-1`? We're not on any `v5-09-*-patches` branch.

TimoWilken:
Not needed. I changed the alidist commit to only change AliRoot's version number, and to build with the then-current AliPhysics release, so the testing release was published anyway.

PR #1035, opened 2021-06-29
Move remaining plain slc8 builders to cs8

TimoWilken:
The plain cs8 image is working, so switch the remaining builders over.

Move or create an `o2` and `o2-dataflow` check each for QC, O2 and alidist, as requested at <https://alice.its.cern.ch/jira/browse/O2-2377>.


ktf:
Will this use the s3 backend? I do not want everyone to wait for 12 hours to rebuild stuff on the new builders.

TimoWilken:
Locally, in cs8-builder, everything in an O2 build is fetched from the remote except for O2 itself (which would have to be rebuilt for each PR anyway).

ktf:
Ok. Let's see.

PR #1041, opened 2021-07-05
Add script to show an overview of recent PRs

TimoWilken:
The script queries recent pull requests' CI statuses for each known check and displays the results in a concise and colourful table for easy analysis.

Also extract `.env` file parsing into a library function as it's used by both `list-branch-pr` and the new script.

The new script uses GraphQL, so add a dependency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ python-ldap[services]
 boto3
 s3cmd
 pyyaml
+gql>=3.0.0a6
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,4 @@ python-ldap[services]
 boto3
 s3cmd
 pyyaml
-gql>=3.0.0a6
+gql==3.0.0a6; python_version >= '3.6'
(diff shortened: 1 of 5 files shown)


TimoWilken:
Draft because the PR checker script seems to have trouble installing this.

requirements.txt
```diff
@@ -7,3 +7,4 @@ python-ldap
 boto3
 s3cmd
 pyyaml
+gql>=3.0.0a6
```
ktf, line 10: Please hardcode a version.

TimoWilken:
The installation on python2 is fixed, though `ci-status-overview` (and anything else that wants boto3 or gql -- so `aliPublishS3`, `repo-s3-cleanup`, `report-pr-errors` and `update-symlink-manifests`) will fail with an `ImportError` if run under python2. We're running all of those in python3 environments, though.

ktf:
Would it be possible to run the script and produce a summary page automatically on every PR which is merged? We could then populate some page in aliceo2group.github.io with the results.

PR #1043, opened 2021-07-05
Reduce number of jobs for mesosdaq

TimoWilken:
QC PRs have out-of-memory errors, both in the O2 build and in some QC tests. 12 is too much, let's see if 10 works; hopefully we don't have to reduce it to 8 again.


ktf:
Can you point me to one?

TimoWilken:
https://github.com/AliceO2Group/QualityControl/pull/757

ktf:
Is this correlated with bringing back alientest06 in business?

TimoWilken:
I was mistaken earlier -- it was actually alientest02 where I changed the way mesos was run recently. Maybe taking mesos offline on alientest02 meant that one or two extra slc7 builders were scheduled on alientest06 and are taking up memory there, I'm not sure.

ktf:
that should not happen, in the sense that mesos should avoid the resource overallocation. Can you check how many zombies you have on that machine?

TimoWilken:
Mesos doesn't actually check that builders are only using their assigned 24G of memory and no more, IIRC. So if multiple builders go over that at the same time, that could cause problems.

There are quite a lot of zombies in some of the containers on that machine. I could restart those workers.

ktf:
No, mesos / docker will actually enforce that. That's why they fail. It's not the global limit that we are passing, but the one in the container.

ktf:
Regarding the zombies, in principle your cs8 changes should eliminate them, can we move to the new way of building stuff also for CC7?

PR #1044, opened 2021-07-06
Post to Mattermost when new RPMs are published

TimoWilken:
Tested and working.


TimoWilken:
Note to self: when this is merged, I need to `aurora update start` both running RPM creation tasks.

PR #1045, opened 2021-07-07
daily-tags: Don't take up an slc7-light node when it's not needed

TimoWilken:
Currently, the Jenkins daily-tags script takes up a slot on an slc7-light node for the full duration of the build, even though that slot is only needed for the initial step of waiting for PRs to be merged (and most daily builds don't even use that step).

With this patch, that slot is only used as long as it is needed, and is freed when the real build begins on another node.

The diff contains mostly whitespace changes.

### DIFF
--- a/jenkins/daily-tags
+++ b/jenkins/daily-tags
@@ -1,22 +1,22 @@
 #!groovy
-node ("slc7_x86-64-light") {
-  try {
-    stage('Construct tag name') {
-      /* Turn AUTOTAG_PATTERN into AUTOTAG_TAG now, before waiting for PRs,
-       * because waiting can take hours, and that would mean we use the incorrect
-       * date to name the tag.
-       * Run on any free node -- doesn't matter which, but we need one to use sh.
-       */
-      node {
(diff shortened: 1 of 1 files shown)


ktf:
This is fine, my only worry is that this might be introducing some unexpected parallelism in the builds.

TimoWilken:
What sort of parallelism do you mean? Within a build or between builds?

Within a build, steps are executed in order, whether or not they're running on a particular node. So the build would still have to wait for the PR stage to complete.

As for multiple builds running at the same time: currently (without this patch), each build takes up one slot on the slc7-light builder, of which there are 4. So at the moment, multiple daily-tags invocations can already run at the same time. If those are all full, Jenkins provisions ad-hoc slc7-light nodes via mesos, so even those 4 slots aren't really a hard limit. (With this patch, it probably wouldn't have to do that at all, which cuts down on the number of things running under mesos that we don't control via aurora.)

If the first stage (waiting for PRs) completes, but there is no build node available, the behaviour doesn't change with this PR merged -- the build still goes in the queue and waits for one to become available, just like now. The difference is that while it's in the queue (and during the subsequent build), it won't needlessly take up an slc7-light spot.

Am I missing something?

ktf (approved):

PR #1049, opened 2021-07-16
Publisher: Install concurrent.futures for boto3

TimoWilken:
Boto3 is dropping support for python2, so we should move aliPublish over to python3. As a stopgap to make the publisher work for now, install concurrent.futures ourselves.


ktf:
Can't we fix setup.py to require concurrent.futures (and boto3) instead? Or can't we simply hardcode the version of boto3 to one which works for python2?

PR #1050, opened 2021-07-21
Move to python3 for aliPublishS3

TimoWilken:
Tested and working for the el8 publisher.


TimoWilken:
The scripts are working and publishing properly now.

PR #1053, opened 2021-07-28
Move O2Physics check to a dedicated builder

TimoWilken:
Draft while I create the builder.


TimoWilken:
The separate builder is running now, ready to merge.

PR #1056, opened 2021-08-05
Run aliPublishS3 off S3 repository only

TimoWilken:
With this patch, aliPublishS3 won't need a full local copy of the entire repo any more. Instead, it will merge new RPMs into the S3 repo directly, and upload the RPMs directly as well (so the `rclone sync` step isn't needed for aliPublishS3 any more).

Publishing tested and working using the `el8.test` architecture.

### DIFF
--- a/publish/aliPublishS3
+++ b/publish/aliPublishS3
@@ -3,13 +3,11 @@
 import logging, sys, json, yaml, errno, boto3, requests
 from glob import glob
 from argparse import ArgumentParser
-from datetime import datetime
 from time import time
 from logging import debug, info, warning, error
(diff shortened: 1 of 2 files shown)


ktf (commented):
Ok to merge, but let's check with @ironmann if it works for him.

ktf (changes requested):
Please fix the merge conflict.

TimoWilken:
Unfortunately, this change means that the code that posts new packages in Mattermost will stop working for el8 (as that code uses the on-disk RPMs). That should be easy to fix, but is probably better done in a new PR.

PR #1057, opened 2021-08-13
Install dependencies for aliPublish

TimoWilken:
aliPublish is run using python3, so use pip3 to install.


TimoWilken:
@ktf: untested! I don't know if `--user` is the right thing to do here.

PR #1060, opened 2021-08-15
Only install minimal dependencies for the publisher

ktf:
Apparently using requirements.txt forces full installation, including the [services] stuff.


ktf:
@TimoWilken

PR #1061, opened 2021-08-16
Make aliPublish work with python3 for alien

ktf:


ktf:
@costing once this is merged, can we have a look if it still works on alimonitor?

PR #1062, opened 2021-08-17
Only process one PR at a time

TimoWilken:
This cuts down on long queues in CI builders, so they report pending statuses more often.

### DIFF
--- a/list-branch-pr
+++ b/list-branch-pr
@@ -286,7 +286,7 @@ def main(args):
 
     if grouped["untested"]:
         # If there are untested PRs waiting, build all of them first.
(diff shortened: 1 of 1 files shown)


ktf:
Do you sort the hashes by date? we want to avoid that something which is 2nd suddenly becomes 10th.

ktf:
I guess we should actually do this. Right now, if someone pushes two commits, we will keep building the old commits because we do not check early enough. This should actually help reduce the build time although the queue position might start to jump. @TimoWilken how difficult would it be to build the list according to the time of the last commit?

PR #1063, opened 2021-08-17
Don't change defaults-*.sh in daily tags script

TimoWilken:
If we have the O2 version number just in o2.sh, we don't need to change the defaults as well.

Not tested.


ktf:
This is fine, but did you verify the rebuild happens for any change to defaults-o2.sh?

ktf:
Thinking again about it, if the current behaviour is, at it seems, not to add a forced version if the associated override stanza is not there, then the behaviour can stay as it is, I guess...

PR #1065, opened 2021-08-18
Only enable O2 tests when building O2 or alidist

TimoWilken:
This removes the blanket `ALIBUILD_O2_TESTS=1` setting in CI and only enables O2 tests if PR_REPO is O2 or alidist. The fullCI builders are an exception, as they explicitly disabled O2 tests before, so they are kept disabled.

@ktf, is this OK? Did I miss something? Or should we keep the system where O2 tests are enabled by default and only disabled in some cases?


ktf:
I would make enabling tests opt in, so this is fine.

PR #1066, opened 2021-08-19
Make daily-tags compilation jobs configurable

TimoWilken:
Now that O2Physics is a separate repo, it and O2 shouldn't take as much memory to compile.


daily-tags.sh
```diff
@@ -132,7 +132,7 @@ aliBuild --reference-sources mirror                    \
          --debug                                       \
          --work-dir "$workarea"                        \
          ${ARCHITECTURE:+--architecture $ARCHITECTURE} \
-         --jobs 8                                      \
+         --jobs 12                                     \
```
ktf, line 135: This is fine, but don't we control that with a JENKINS variable as well? Maybe it should be $JOBS, no?

PR #1070, opened 2021-08-20
Use python3 for aliBuild also on cc7

ktf:
This will allow us to drop python2 builds on CC7.

### DIFF
--- a/.github/workflows/o2-full-deps.yml
+++ b/.github/workflows/o2-full-deps.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     container: centos:7
     env:
-      ALIBUILD_TAG: v1.8.8
+      ALIBUILD_TAG: v1.8.9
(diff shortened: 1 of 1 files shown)


ktf:
@awegrzyn @Barthelemy, heads up: this will make sure that aliBuild gets installed using python3 rather than python2.

This will allow us do get rid of the python2 O2 nightly builds which are currently causing every nightly to have -1 and -2 revisions. Besides the wasted resources, this is confusing people.

The drawback will be that if you explicitly install using python2 (e.g. explicitly doing so via pip rather than via the rpms), you will not have cached tarballs anymore. Can you confirm me this will not affect you? If it does, could you switch to https://alice-doc.github.io/alice-analysis-tutorial/building/prereq-centos7.html to install alibuild? It should be in any case a more sustainable approach, compared to using PIP.

awegrzyn:
Hi, yes I think that's fine, as far as I know most of the people already migrated to CS8 (or using macOS).

PR #1072, opened 2021-08-20
Parallel tagging of related packages

TimoWilken:
Allow tagging multiple packages with the same tag in daily-tags.sh.

This is needed to tag O2, then tag and build O2Physics with the same tag immediately afterwards.

### DIFF
--- a/daily-tags.sh
+++ b/daily-tags.sh
@@ -1,11 +1,17 @@
 #!/bin/bash -ex
-set -x
+set -ex
 
 # Check for required variables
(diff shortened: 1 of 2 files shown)


TimoWilken:
Hi @ktf -- sorry for the delay, but this seems to work now. The last test run produced the following alidist commit for a parallel O2/O2Physics tag: https://github.com/alisw/alidist/commit/af49509ca95ce5b0614c72448a7fe2562c15bc69.

The Jenkins daily builds will need updating when this is merged, though -- could we wait until tomorrow morning?

ktf:
yes, sure. Let's have a look tomorrow.

PR #1073, opened 2021-08-30
Update o2physics.env

ktf:


ktf:
@TimoWilken I think we can move to 12 cores now, no? I think that's what we use for the daily builds in any case, no?

TimoWilken (approved):
Yes, the daily builds use 12 cores.

PR #1075, opened 2021-08-31
daily tags: Don't set tag/version overrides if not needed

TimoWilken:
If the defaults file doesn't contain a `tag:` or a `version:` override for the packages we're building, it is not changed. If it contains either, the specified override(s) are forced to the tag/version we're building.

Untested, but it's a fairly simple change.


ktf:
How can we test this?

TimoWilken:
I could run a DailyO2-testing job for O2 (which has no overrides in `defaults-o2.sh`) and AliPhysics (which does in `defaults-next-root6.sh`), and see if it does the right thing?

ktf:
Yes, please.

PR #1078, opened 2021-09-07
Increase number of parallel builds in CI

TimoWilken:
Now that O2Physics is split out of O2, memory use should go down a lot. Use `nproc` JOBS on MacOS (which is 6 on `alibuildmac03`), and the regular builders' `nproc` of 16 on Linux. (The `alientest`s have more cores, but we shouldn't use all of them at once.)


ktf:
Merging this now that we should have memory usage under control. In any case we should reduce the number of build processes using the ninja "pool" feature rather than globally.

PR #1084, opened 2021-09-14
Publish O2sim

ktf:


ktf:
In case we prefer to have it separate from O2PDPSuite

PR #1085, opened 2021-09-20
DPL: allow overriding recipes with those in a given branch

ktf:


ktf:
@davidrohr as discussed. CC @ironMann

PR #1089, opened 2021-09-20
Use O2Physics master for the CI of O2

ktf:


ktf:
@jgrosseo as we discussed. This should make O2Physics master tested rather than the already built version.

jgrosseo:
Thanks. When it is in, I can merge the next synced update.

PR #1091, opened 2021-09-30
Add "FLPSUITE_LATEST" variable to ALIDIST_SLUG for daily tags

TimoWilken:
The FLP suite needs special handling, so add a variable (which can be used when scheduling a build in Jenkins), which is substituted for the current FLP suite version.

### DIFF
--- a/daily-tags.sh
+++ b/daily-tags.sh
@@ -3,6 +3,22 @@ set -ex
 
 # Check for required variables
 ALIDIST_SLUG=${ALIDIST_SLUG:-alisw/alidist@master}
+if echo "$ALIDIST_SLUG" | grep -q '!!FLPSUITE_LATEST!!'; then
+  case $(date +%u) in
+    1)  # Monday (for Sunday night build)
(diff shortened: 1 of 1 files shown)


davidrohr (approved):

daily-tags.sh
```diff
@@ -3,6 +3,22 @@ set -ex
 
 # Check for required variables
 ALIDIST_SLUG=${ALIDIST_SLUG:-alisw/alidist@master}
+if echo "$ALIDIST_SLUG" | grep -q '!!FLPSUITE_LATEST!!'; then
```
ktf, line 6: I would just use FLPSUITE_LATEST. Why the extra `!!`?

TimoWilken: To make it obvious that it'll be replaced and isn't just a normal tag.

ktf: Well, it would have been good enough. Anyways, fine. Merging this. Can you update the jenkins job accordingly?

ktf: (I mean DailyO2Dataflow)

PR #1102, opened 2021-10-22
Look for newly published RPMs on S3 instead of locally

TimoWilken:
This makes the Mattermost notification work on architectures where we only publish to S3, not locally.

Tested and working using the `el8test.x86_64` architecture.


ktf:
Pilot beam is over. Let's merge this and keep an eye on whether you still see the messages.

PR #1103, opened 2021-10-26
Post comments in PRs with CI errors

TimoWilken:
Temporary measure until a better solution is implemented, as discussed at WP3.

For now, applies to PRs in the AliceO2Group/AliceO2 and AliceO2Group/QualityControl repos.

Cc: @chiarazampolli @Barthelemy


ktf:
Ok, now things seems to be green everywhere. I will wait one day to make sure that's actually the case and then we can merge this.

PR #1104, opened 2021-11-01
Don't use pip install --ignore-installed

TimoWilken:
This causes errors when installing a different version of alibuild and then trying to run it.


ktf:
Mmm... maybe it would be a good idea to uninstall alibuild, before installing? We might want to downgrade the current installation...

TimoWilken:
It seems that pip already handles this case -- if I do, for example, `pip3 install alibuild && pip3 install -U alibuild==1.8.2`, it automatically uninstalls 1.9.1 to install 1.8.2, and I end up with aliBuild v1.8.2.

ktf:
Ok, let's keep an eye on this.

PR #1111, opened 2021-11-03
Revert "Post comments in PRs with CI errors"

ktf:
Reverts alisw/ali-bot#1103


ktf:
Two issues:

* It looks in the wrong file for reporting.
* We still have spurious issues with CCDB and QC which we should understand before merging.

PR #1112, opened 2021-11-04
Add CI checks on new M1 Mac

TimoWilken:
Add a check for O2, alidist, QC and O2Physics.

Draft until the builder is fully deployed.

Cc: @MichaelLettrich


TimoWilken:
I've changed this to only add an alidist check. On the machine itself, I've added a `config/silent` so it doesn't upload results.

PR #1119, opened 2021-11-12
aliPublishS3: Check for tarball before creating RPM

TimoWilken:
alibuild uploads the tarball last, so if it doesn't exist yet, the list of symlinks in dist/ that we got from S3 is likely incomplete. Therefore, skip publishing the package for now.

Untested.


TimoWilken:
Tested and working.

ktf (approved):

PR #1120, opened 2021-11-18
Recognise coloured error messages

TimoWilken:
Also, remove `^fatal:` messages -- these are from git, and are mostly false positives.

### DIFF
--- a/report-pr-errors
+++ b/report-pr-errors
@@ -31,7 +31,7 @@ except ImportError:
 
 
 ERRORS_RE = re.compile(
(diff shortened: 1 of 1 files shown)


ktf:
How can "^fatal:" messages not be worrisome?

TimoWilken:
For some reason, some packages' `configure` steps call `git` assuming they're in a git repo (which they aren't). This is presumably optional, as the build doesn't fail, but it means that unrelated logs sometimes have a bunch of `fatal: Not a git repository` messages in them.

PR #1122, opened 2021-11-22
Revert "Publish FOCAL on slc7 as well"

ktf:
Reverts alisw/ali-bot#1121


ktf:
Why was it not build on slc6?

PR #1133, opened 2021-12-16
Use O2Physics-specific CI variable

TimoWilken:
...to avoid running O2 tests in O2Physics CI.

This enables O2Physics' `-Werror` for O2, O2Physics and alidist CI (except fullCI, which didn't have `ALIBUILD_O2_TESTS`). Is this correct?


ci/repo-config/mesosci/cs8/o2-alidist-dataflow.env
```diff
@@ -11,3 +11,4 @@ DEVEL_PKGS="$PR_REPO $PR_BRANCH
 AliceO2Group/AliceO2 dev O2
 AliceO2Group/O2Physics master"
 ALIBUILD_O2_TESTS=1
+ALIBUILD_O2PHYSICS_TESTS=1
```
ktf, line 14: For dataflow we probably do not need the O2PHYSICS tests...

PR #1138, opened 2022-02-09
Update check message when rebuilding

TimoWilken:
When we're automatically rebuilding a check, it would be helpful for users to have feedback about this, so change the existing status's message (keeping the red/green status and results URL intact).

This requires an extra option to set-github-status to keep the URL intact, which is added by this commit as well.

Untested so far, so I've made this a draft PR for now.


ktf:
Looks like a good idea to me.

ktf:
Is this still draft?

ktf:
Ping?

TimoWilken:
Tested and working, as far as I can tell:

Screenshot of new in-progress status

OK to merge.

PR #1139, opened 2022-02-09
Give anyone with parent directory rights subdir rights as well

TimoWilken:
E.g. if `@oadb` is allowed access to `OADB/`, then give them access to `OADB/PWGLF/` as well, even if the latter directory is covered by another rule.

The point about `num_approve` not being handled entirely correctly may be moot, as I can't find any config actually using that setting.

@ktf, @pzhristov: do you think this implementation satisfies the requirements to allow access to subdirs?

## Examples

For example, the new implementation would newly allow access to:

- `OADB/{PHOS,PWG,PWGCF,PWGDQ,PWGGA,PWGHF,EMCAL,PWGPP,PWGUD}/` to `@oadb`
- `PWGCF/FEMTOSCOPY/{FemtoDream,macros}/` to `@alice-svn-pwgcf`
- `PWGPP/{ITS,benchmark,EMCAL}/` to `@alice-svn-pwgpp` and similar for other `PW*` top-level dirs
- everything to `@aliroot-git-admins`
- and more

Is this OK?

### DIFF
--- a/ci/process-pull-request-http.py
+++ b/ci/process-pull-request-http.py
@@ -123,13 +123,22 @@ class State(object):
     else:
       for fn in pull.get_files():
         debug("determining permissions for file %s" % fn)
+        num_approve_this_file = 0
+        approvers_for_file = set()
         for rule in perms:
           num_approve,approve = rule(fn, pull.who)  # approve can be bool or set (not list)
           if approve:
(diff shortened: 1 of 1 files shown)


pzhristov (approved):
Yes, this is the original idea: those who have access to the top directories should have access also to the subdirectories.

PR #1140, opened 2022-02-15
Fix PR processing: handle case where rule is a bool

TimoWilken:
In some cases, the "list" of approvers can be a boolean instead of a set. That
case wasn't handled by my previous change.


TimoWilken:
Merging as the current behaviour is broken.

PR #1141, opened 2022-03-01
Get an AliEn token and make it available to CI builds

TimoWilken:
- If possible, get a temporary AliEn token and store it in env variables.
- We need an upgraded pip in order to install alienpy, but that breaks
  installation of ali-bot and alibuild on some platforms (e.g. Ubuntu).
  Installing them as editable solves this, but needs an `#egg=` fragment at the
  end of the URLs.

Tested and apparently working.

### DIFF
--- a/ci/build-loop.sh
+++ b/ci/build-loop.sh
@@ -159,6 +159,14 @@ fi
 build_identifier=${NO_ASSUME_CONSISTENT_EXTERNALS:+${PR_NUMBER//-/_}}
 : "${build_identifier:=${CHECK_NAME//\//_}}"
 
+# Get a temporary JAliEn token certificate and key, to give anything we build
+# below access.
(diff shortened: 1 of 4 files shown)


ci/build-loop.sh
```diff
@@ -159,6 +159,14 @@ fi
 build_identifier=${NO_ASSUME_CONSISTENT_EXTERNALS:+${PR_NUMBER//-/_}}
 : "${build_identifier:=${CHECK_NAME//\//_}}"
 
+# Get a temporary JAliEn token certificate and key, to give anything we build
+# below access.
+if jalien_token=$(alien.py token -v 1); then
```
ktf, line 164: Is this a read only token?

TimoWilken: No idea, presumably not.

ktf:
Is this ready to go?

PR #1142, opened 2022-03-05
Drop o2-dataflow builds

ktf:
CS8 is what matters now.


ktf:
@TimoWilken is this all it is needed?

TimoWilken:
Depends if you only want to get rid of it for alidist or not. There's also <https://github.com/alisw/ali-bot/blob/master/ci/repo-config/mesosci/slc7/o2suite-dataflow.env> for AliceO2Group/AliceO2.

The deleted checks also need to be removed from the list of required checks in the repo's branch protection rules.

PR #1144, opened 2022-03-16
Deduplicate warning messages

TimoWilken:
For compiler warnings, we don't need the context lines, and compiler warning messages might be repeated many times, so filter out duplicates.


TimoWilken:
Found an error I need to fix before merging this.

ktf:
Is this ready to go?

ktf:
What is the status  of this one?

ktf:
Ping?

PR #1145, opened 2022-03-25
Fix FLP suite version used in daily tags

TimoWilken:
On Mondays, fetch the latest available tag; else, build against the installed version.

Cc: @davidrohr


TimoWilken:
Merged and tested by rebuilding last night's build. At least for that, it picked up the right tag.

davidrohr:
thx, seems to work. Today's build was correct. Let's also check Monday's build.

PR #1146, opened 2022-03-25
Change list-branch-pr to use GraphQL

TimoWilken:
This uses much fewer GitHub API requests than the script relying on the REST API, so it should help us stay within quota.

Python-gql 3.1.0 has been released, so we don't need to keep pinning a prerelease version any more.

Still needs testing, so draft for now.


TimoWilken:
I've tested this on two CI builders, and it works fine. Ready to merge from my point of view.

PR #1148, opened 2022-03-31
Use alibuild v1.10.2 in CI

TimoWilken:


ktf:
Another tag needed in any case.

PR #1150, opened 2022-04-07
Rework daily tags script

TimoWilken:
- allow overriding arbitrary package's version
- provide placeholders for arbitrary packages' version
- provide placeholder for both latest and current flpsuite

Cc: @davidrohr


davidrohr (dismissed):
thx, without checking the code in detail, this looks feature-wise like what we want.

ktf:
As discussed privately, rather than trying to replicate the parsing of the alidist recipes you would be better off writing a simple build plugin for alibuild which uses the same mechanics aliBuild build is using and simply prints out a formatted string.

davidrohr:
I agree with @ktf , but if you assume this takes few days, I'd prefer if we could merge this as is (assuming it works), and then clean it up in a second PR.

ktf:
Is this still relevant?

PR #1151, opened 2022-04-11
Use defaults o2 for AliPhysics with root6.

ktf:


ktf:
@pzhristov as discussed today.

PR #1152, opened 2022-04-12
Add a CI check in O2DPG

TimoWilken:
Build and check O2Suite (with O2 from dev and O2Physics from master) for PRs in O2DPG.

As requested at WP3 meeting.


davidrohr:
What package will this test build? we should build o2-full-system-test.sh directly, not o2-fullci.

TimoWilken:
I initially had it as only O2Suite, but I've changed it to O2-full-system-test now.

TimoWilken:
In summary: the check will build O2-full-system-test with `o2` defaults against O2@dev and O2Physics@master, including their respective unit tests.

davidrohr:
@TimoWilken : Sorry to bug again. we don't need the unit tests of O2/O2Physics. Can you disable them (Don't set ALIBUILT_O2_TESTS=1, as in the FullCI for O2)

PR #1153, opened 2022-04-12
Implement required changes for O2PDPSuite release process

TimoWilken:
This is a draft of the reworked implementation, as requested by @ktf. This also requires changes to aliBuild.

- allow templating in AUTOTAG_PATTERN (date formatting is done afterwards)
- allow overriding specific packages' tags

Cc: @davidrohr

### DIFF
--- a/daily-tags.sh
+++ b/daily-tags.sh
@@ -1,21 +1,21 @@
 #!/bin/bash -ex
 set -ex
 
-edit_tags () {
-  # Patch package definition (e.g. o2.sh)
-  local package tag=$1 version=$AUTOTAG_OVERRIDE_VERSION
-  for package in $PACKAGES; do
-    sed -E -i.old \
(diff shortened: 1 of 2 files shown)


davidrohr (dismissed):
seems OK in the weekly build from today

TimoWilken:
In that case, I'm happy with it too.

TimoWilken:
716e90bb197d05a742667a14b17d4f82d21d1d49 implements one more feature @ktf asked for, namely that QC should be tagged, but not included in the final alidist commit. This seems like the easiest place to put this as well; I hope that's alright.

ktf:
Are we ready to merge this?

TimoWilken:
Fine by me! I can check it ran correctly tomorrow morning.

ktf:
Do we need to change the jenkins rule for this to work?

TimoWilken:
I don't think so. There are only a few additional env vars, but if they're not declared as parameters, they'll just be unset, which is fine.

PR #1168, opened 2022-07-06
Clean up unused builds after 2 days

TimoWilken:
After every build, delete builds and installations of development packages that haven't been touched in two days. This should hopefully keep disk usage under control.

A better solution would probably be to monitor the absolute disk usage of the CI builder and delete as many old builds as we have to to stay within our limit, but that would be much more complicated, and counting our own disk usage would take a long time.

Also, monitor the time taken by the cleanup process. For this purpose, factor out a function for pushing to InfluxDB, which could be used to push additional metrics in the future.

Untested so far.

### DIFF
--- a/ci/build-helpers.sh
+++ b/ci/build-helpers.sh
@@ -3,6 +3,28 @@
 # It is sourced on every iteration, so functions defined here can be overridden
 # while the builder is running.
 
+function influxdb_push () {
+  # Usage: influxdb_push TABLE TAG=V TAG=V -- FIELD=V FIELD=V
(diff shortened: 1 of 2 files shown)


ktf:
Thanks. How do you plan to test this?

TimoWilken:
I've put it in `config/install-alibot` for one builder to see if it works.

ktf:
Did it work? Can we merge this now or you think it's still to be verified?

TimoWilken:
There was one more issue with the new metrics. I'll update when it fully works!

TimoWilken:
Looks fine to me now; cleanup runs and produces metrics. OK to merge in my opinion.

PR #1169, opened 2022-07-09
Replace Mesos with Consul

ktf:


ktf:
@costing @TimoWilken

TimoWilken:
Not sure this will work on alimonitor as it doesn't call the Consul DNS. There's some hackery for `.mesos` URLs in aliPublish, but we can't use the Consul HTTP API like this at the moment. (We could either change the host's `/etc/resolv.conf`, install Consul on the machine or bind the Consul servers to public IPs on the master nodes, if we really want to use `.consul` hostnames.)

`ali-ci.cern.ch` should be better, right?

ktf:
Ok, I am changing to ali-ci.cern.ch. Indeed they should be exactly the same
for that  part.

Ciao,
Giulio

PR #1172, opened 2022-07-21
Smarter cleanup for CI

TimoWilken:
When disk space runs low, clean up more old builds, so that we always have 100 GiB available before every build.

This replaces the old bash-based cleanup procedure with a more full-featured Python script.

Metrics are tracked per-check and pushed to InfluxDB.


TimoWilken:
I've tested this on a CI builder and it's working.

PR #1174, opened 2022-08-02
Update alibuild to v1.12.0 in CI

TimoWilken:


ktf:
Could you solve the conflicts here?

PR #1177, opened 2022-08-24
Expose CI_NAME and CHECK_NAME

ktf:
The idea is to use them to select which checks to run.

### DIFF
--- a/ci/build-loop.sh
+++ b/ci/build-loop.sh
@@ -14,6 +14,8 @@ ensure_vars CI_NAME CHECK_NAME PR_REPO PR_BRANCH PACKAGE ALIBUILD_DEFAULTS
 # Needed so that we can run different kind of tests based on the context
 export ALIBOT_CI_NAME=$CI_NAME
 export ALIBOT_CHECK_NAME=$CHECK_NAME
+export ALIBOT_PR_REPO=$PR_REPO
+export ALIBOT_PR_BRANCH=$PR_BRANCH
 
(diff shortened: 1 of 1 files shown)


ktf:
@TimoWilken what do you think of this?
@Barthelemy this is what we discussed on mattermost.

TimoWilken:
I suppose this ought to work. Where would these vars be used -- in alidist recipes?

Barthelemy:
yes in alidist recipes. In the QC in particular I would like that some tests are only run in the tests in the QC repo and not in alidist.

TimoWilken:
Fair enough. Would exporting `PR_REPO` be a better fit in that case? That always contains the repo name in `organization/repository` format that we are running the check for, whereas `CI_NAME` and `CHECK_NAME` are intended to be more human-readable and can be a bit more free-form.

ktf:
Ok, let's do the full monty, then... Maybe we could switch off some tests based on the branch, for the future where we have both dev and master.

TimoWilken:
One thing I forgot -- you should give these vars to `aliBuild build` explicitly using an `-e "ALIBOT_PR_REPO=$PR_REPO"` option for each one, otherwise they won't be propagated into the build Docker container. Then you also don't need the `export ...` lines.

ktf:
Updated with @TimoWilken suggestions.

Barthelemy:
could we go ahead with this ?

ktf:
@TimoWilken could you redeploy the builders with this, at your convenience?

TimoWilken:
The builders auto-update to the master branch of ali-bot, so it'll happen gradually as they finish their in-progress builds.

PR #1186, opened 2022-09-15
Publish AliPhysics and FOCAL releases on slc7

TimoWilken:
The AliPhysics nightlies are already all built on slc7, so migrate release builds over as well.

@ktf is this OK or should AliPhysics releases stay on slc6?


TimoWilken:
We still need to build on slc6 for now. Closing, pending a proper migration to slc7.

PR #1190, opened 2022-09-21
Speed-up aliPublishS3

TimoWilken:
Fetch directory contents individually from S3, instead of fetching the whole file tree in advance and caching it. This is much quicker if only publishing a few packages.


TimoWilken:
@ktf all done now. I've tested this in the CVMFS publisher and it seems to work, and is a lot faster than before.

PR #1191, opened 2022-09-22
Don't delay daily tags for draft PRs

TimoWilken:
Draft PRs are not tested until marked as non-draft, so there's no point in waiting for them.


TimoWilken:
OK, seems to be working with the AliPhysics dailies. However, I've had to update to a newer PyGithub version (the older one we were using before doesn't expose a `pull.draft` attribute, apparently). The new version isn't available for Python 2, but do we really want to support running `ali-bot` stuff under Python 2?

ktf:
No, we don't especially now that all new builds are done on cc7.

TimoWilken:
@ktf CI is green now. Is this OK to merge?

setup.py
```diff
@@ -15,7 +15,7 @@
 with open(path.join(here, 'README.rst'), encoding='utf-8') as f:
     long_description = f.read()
 
-install_requires = ['PyGithub==1.45', 'argparse', 'requests', 'pytz', 's3cmd',
+install_requires = ['PyGithub==1.55', 'argparse', 'requests', 'pytz', 's3cmd',
```
ktf, line 18: are you sure this works everywhere?

PR #1196, opened 2022-09-28
Publish jemalloc on slc7 as well

TimoWilken:
This is needed to run analysis trains using AliPhysics versions compiled on slc7.


ktf:
@costing We should probably make jemalloc a proper dependency of AliPhysics / O2, if needed.

PR #1200, opened 2022-10-03
Do not quote expression

ktf:


ktf:
@TimoWilken without  this I get an error on mac.

PR #1205, opened 2022-10-12
Run CI builds inside our Nomad-assigned cgroup

TimoWilken:
This should allow Nomad to clean up processes and track CPU use accurately.

Draft as I haven't tested it yet.

### DIFF
--- a/ci/build-loop.sh
+++ b/ci/build-loop.sh
@@ -131,6 +131,8 @@ fi
 unset certdir jalien_token
 
 report_state pr_processing
+PR_OK=0   # If we exit before the end of the script, we didn't succeed.
+trap -- 'report_state pr_processing_done' EXIT
 
 # Fetch the PR's changes to the git repository.
 if pushd "$PR_REPO_CHECKOUT"; then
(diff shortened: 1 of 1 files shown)


ci/build-loop.sh
```diff
@@ -173,6 +175,13 @@ if pushd "$PR_REPO_CHECKOUT"; then
   popd
 fi
 
+# Nomad runs this script inside a cgroup managed by it, so it can clean up
+# properly when we exit (or are killed), and it can track CPU/RAM usage.
+# Run our Docker builds inside the same cgroup so they're included too.
+if cgroup=$(sed -rn '/:freezer:/{s/.*:freezer:(.*)/\1/p;q}' "/proc/$$/cgroup"); then
```
ktf, line 181: Isn't this linux specific?

TimoWilken: Yes it is, that's why it's inside the `if`.

On Mac, the `sed` command will fail and we won't change `DOCKER_EXTRA_ARGS` (though of course, we don't run in Docker on Mac anyway).

TimoWilken:
Tested and working on Linux; this does not affect MacOS at all (the variable we set is not used there). With this change, Nomad now tracks our CPU and memory use during the build as part of the allocation correctly.

PR #1207, opened 2022-10-19
Clean up running CI processes when killed

TimoWilken:
This should make restarting CI processes on Mac much more reliable, as Nomad cannot reliably kill the whole tree itself there. This should also solve the annoyance of CI jobs leaving behind running build containers when they are killed.

The behaviour relies on any long-running processes in the CI scripts being wrapped in a short_timeout or long_timeout call, which I think they all are.

Draft because I've not tested this yet.


TimoWilken:
So far, this doesn't fail when run in normal operation on Linux, but I want to test this change on Mac (and test the behaviour when killed, as well, not just normal operation) before this can be merged.

TimoWilken:
Found a problem with this on the M1 Mac: `wait` returns 0 even if the `aliBuild` process that was put into the background earlier exits with an error.

PR #1210, opened 2022-10-21
Fix overloading of some variables in `.env` files for CI

TimoWilken:
For some permissions-related variables (such as `TRUST_COLLABORATORS`), this fixes buggy behaviour where the most-permissive value encountered would be applied, even if overridden in a more-specific `.env` file.

As far as I can tell, this problem only occurred with the Mac CI check in AliRoot, where the check would run for anyone who had authored commits in the AliRoot repo, instead of only after a `+1` comment.


TimoWilken:
Tested and working; merging.

PR #1221, opened 2022-11-18
Support non-x86_64 architectures in alienv

TimoWilken:
This is required for aarch64 support.

The hardcoded x86_64 string was not changed in `load_alien_if_missing`, which should only run on `el5` -- and we don't have any `el5` ARM machines.


TimoWilken:
Changes applied on CVMFS; merging this to keep the repo in sync.

PR #1226, opened 2023-01-11
Add timeout for alien.py call

TimoWilken:
This command was found hanging on one builder, stopping it from building PRs.


ktf:
Thanks. @costing is that expected to hang?

PR #1227, opened 2023-01-20
Fix code paths for updatable RPMs

TimoWilken:
This branch includes the changes to the CentOS 8 RPM publisher that I've been testing recently. Most importantly, this allows us to publish updatable RPMs for CentOS 8 (i.e. off S3).

At David's request, I also changed the updatable RPM publishing script to install stuff in package-specific (but not version-specific) subdirs, so e.g. O2 executables would go in `/opt/alisw/elX/O2/bin/`, instead of `/opt/alisw/el8/bin/`. Modulefiles still all go in `/opt/alisw/elX/etc/Modules/modulefiles/`.

Tested and working as far as I can tell, but comments/feedback would be appreciated.

### DIFF
--- a/publish/aliPublish-rpms-cc8test.conf
+++ b/publish/aliPublish-rpms-cc8test.conf
@@ -1,0 +1,68 @@
+# vim: set filetype=yaml:
+---
+s3_endpoint_url: https://s3.cern.ch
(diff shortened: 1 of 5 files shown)


TimoWilken:
Merged, since this has been running in production without problems for weeks now.

PR #1228, opened 2023-02-01
Sync clang-format version with the alidist recipe

vkucera:
Related PR in alidist: https://github.com/alisw/alidist/pull/4791


TimoWilken:
Thanks Vit!

PR #1230, opened 2023-02-09
Track number of commits on the base branch in CI

TimoWilken:
This metric would be useful to see which commits on the dev/master branch introduce regressions.


ktf:
Looks good to me.

PR #1233, opened 2023-03-07
Track commit push time

TimoWilken:
This should give a good approximation of the time a commit was submitted for testing. It still breaks down for draft PRs that are marked non-draft after a while.

Draft because it needs testing.


TimoWilken:
Tested and working in CI now @ktf; ready for review!

ktf:
As long as the only side effect is that draft PRs will not have the correct number associated, I think we do not care.

TimoWilken:
Yes, the problem only occurs for PRs that were draft at the time of their latest push (or creation) and are then made non-draft later. In that case, I'll merge this PR, since I've tested the changes already.

PR #1248, opened 2023-06-02
Extend symlink caching script to handle dist symlinks

TimoWilken:
This should make publishing much faster, once publishers can start to use pre-built manifests to avoid having to walk the full dist tree on S3 themselves.

For now, nothing uses these new dist manifests yet; the next step would be to make `aliPublishS3` read them.


TimoWilken:
Tested and appears to work locally; nothing uses the new manifests yet anyway, so merging.

PR #1250, opened 2023-06-14
Run AliEn publisher under Nomad using authenticated alimonitor API

TimoWilken:
This allows us to run the AliEn publisher elsewhere, e.g. through Nomad for better monitoring.

Draft and untested since the auth mechanism isn't implemented yet by @costing as far as I know.

### DIFF
--- a/publish/aliPublishS3
+++ b/publish/aliPublishS3
@@ -186,10 +186,16 @@ class CvmfsServer(PlainFilesystem):
 class AliEn:
 
   def __init__(self, connParams, dryRun=False):
-    self._req_params = {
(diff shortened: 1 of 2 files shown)


ktf:
For me this is a more than welcome change. @costing do you have any estimate on your side on the ETA for the API to be production ready?

TimoWilken:
I've tested the basic functionality, except actually registering a new package with AliEn. For that, I'd wait until tomorrow morning, when the daily tags are published.

costing:
The API is ready, let's give it an exclusive try and see how that goes.

PR #1251, opened 2023-06-14
Change the default remote store to S3

TimoWilken:
The rsync repo is obsolete and everything should upload directly to S3 instead.


TimoWilken:
I've separately changed all daily Jenkins jobs' default `REMOTE_STORE` parameter to the `b3://...` one, but this changes the default (so when `REMOTE_STORE` is empty) and removes the footgun where some things would upload to the defunct rsync repo in that case.

PR #1257, opened 2023-07-05
Handle 2X.* versions of Ubuntu in alienv

TimoWilken:
Might be best to just use a wildcard. If you prefer explicitly listing version numbers instead, let me know @ktf.


TimoWilken:
Merged since the test still fails for mysterious reasons, but it can run the `alienv` script fine.

TimoWilken:
...and published on the CVMFS Stratum-0 @zensanp! Should propagate within half an hour or so to your instance.

PR #1258, opened 2023-07-10
Handle distro ID on RHEL 9

TimoWilken:
It now contains spaces.


TimoWilken:
The test still fails as before, but the first few checks are fine, so no problem with the syntax. Merging.

PR #1261, opened 2023-08-23
Add pyproject

ktf:
Apparently this is now required to be able to do pip install from git


ktf:
Newer versions of pip hang if this is not there.

PR #1267, opened 2023-11-01
Gradually re-index RPMs if they are missing from the repo index

TimoWilken:
Change `aliPublishS3` for RPMs to download the latest package index file, and if it finds any uploaded `.rpm` files that are not in the index, download up to ~10GiB of them and add them to the index.

This will gradually rebuild the index to include all uploaded RPMs.

Also, backup old index files before replacing them (so we never catastropically lose the entire index), cleaning up index files that are older than a month.

Tested and working with the el8 RPM repo.

### DIFF
--- a/publish/aliPublishS3
+++ b/publish/aliPublishS3
@@ -1,16 +1,19 @@
 #!/usr/bin/env python3
 
-import logging, sys, json, yaml, errno, boto3, requests
+import logging, gzip, sys, json, yaml, errno, boto3, requests
+import botocore.exceptions
 from glob import glob
 from argparse import ArgumentParser
+from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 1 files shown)


ktf:
This is actually a very nice development, thank you. As discussed a separate improvement over this would be to make sure you sort missing packages by their upload date and then prefer their dependencies first. This would potentially reduce the time one has to wait for recent releases in case of major losses in the database.

TimoWilken:
Thanks @ktf. I'll do as you suggest, but in the meantime, would it be alright to merge this? I've already tested this change by running it as the RPM publisher for an afternoon and it has already recovered a bunch of packages, but I reset the publisher to the `master` branch pending the merge of this PR.

PR #1268, opened 2023-11-01
Unify handling of GitHub cache file

TimoWilken:
This makes all scripts using the GitHub REST API use the same cache file, hopefully reducing the number of API requests slightly overall, and leaving fewer cache files all over the disk.


TimoWilken:
Alright, ready now. Sorry for the spam! Tested with `bulk-change-pr-status` and `list-branch-pr`.

PR #1269, opened 2023-11-06
Unify RPM repo creation workflows

TimoWilken:
Deduplicate RPM creation jobs, so that there isn't a copy-pasted one for each CentOS/Alma version.

Configuration is done through a matrix, similarly to the Ubuntu workflow.


ktf (approved):
Feel free to merge if you are happy with it.

PR #1270, opened 2023-11-06
Remove and/or migrate unneeded slc7 dataflow checks

TimoWilken:
o2-dataflow defaults are not used on slc7 anywhere now, so these checks aren't needed.

For checks that already have a cs8 equivalent, delete the slc7 check.

For low-traffic repos which only have a slc7-dataflow check, move it to cs8.

Cc: @Barthelemy


ktf:
Shouldn't we move them to Alma9 directly?

Barthelemy:
Thank you Timo. 
We are still not sure if we move to alma/rhel 9 or not. We will move away from CS8 but perhaps to RHEL 8 for the time being. 
Tests are ongoing and we will also take into account what EPNs do.

ktf (approved):

ktf:
Fair enough.

PR #1271, opened 2023-11-22
Do not depend on bc being installed

ktf:


ktf:
@TimoWilken could you cross check?
@sy-c I guess we do not need the RPMs for DA anymore...

sy-c:
> @sy-c I guess we do not need the RPMs for DA anymore...

I hope so, I have not seen one for the past 5 years ;) please go ahead and cleanup related features as you need !

ktf:
That's the answer I wanted... ;-)

TimoWilken:
Do we need this `BUILD_DATE` calculation anyway?

As far as I can tell, it creates a new build area every 3 days (not sure why), then later in the script it loops through all the build areas to find its own one (the `$WORKAREA/$WORKAREA_INDEX/current_slave` stuff), I guess to deconflict different "build slots"/executors on the same builder (why not use [Jenkins' `$EXECUTOR_NUMBER` variable](https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables)?).

If we want a new build area every once in a while, why not e.g. `date -u '+%Y-%m-%d'` for a new one every day or `date -u '+%Y-W%W'` for once a week?

If you want to keep the logic as-is, I'd remove the extra wrapping `$(echo ...)`:
```bash
BUILD_DATE=2015$(("$(date -u +%s)" / 86400 / 3))
```

ktf:
No, indeed we can probably drop everything.

PR #1275, opened 2023-12-04
Make ALIBUILD_*_HASH variables available in build container

TimoWilken:
These are used by o2checkcode and O2DPG.


TimoWilken:
Merging since the change is pretty trivial.

PR #1276, opened 2023-12-11
Allow overriding version field in Jenkins build

TimoWilken:
This lets users override the version field of arbitrary packages in Jenkins builds. This is primarily needed to override the version field of O2sim in async tags' builds.

### DIFF
--- a/daily-tags.sh
+++ b/daily-tags.sh
@@ -1,18 +1,24 @@
 #!/bin/bash -ex
 set -ex
 
+substitute_vars () {
+  local value=$1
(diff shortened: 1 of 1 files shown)


TimoWilken:
@chiarazampolli Once this is merged, you can use the `OVERRIDE_VERSIONS` field in Jenkins, setting it e.g. to `O2sim=async-%Y%m%d.1` (or whatever seems appropriate to you).

sawenzel (approved):
Thanks a lot!

PR #1278, opened 2023-12-12
Set a status message when commits are untrusted

TimoWilken:
People's first contributions to O2 are not tested automatically for security reasons. This often leads to confusion, so set a message on those commits to let people know that they need a PR approval.


TimoWilken:
Tested locally; merging.

PR #1290, opened 2024-02-15
Make el9 the new default platform for CVMFS, falling back as before

TimoWilken:
If we want to build and publish analysis tags for Alma 9 by default, alienv needs to support this platform.

For new tags that are present for el9, use those by default; otherwise, fall back to el7 and other platforms as before.


TimoWilken:
Merged and deployed to CVMFS.

PR #1294, opened 2024-02-23
[WF] Revise message for how to add async labels

benedikt-voelkel:


benedikt-voelkel:
@TimoWilken a tiny necessary correction of the messages that are issued when a PR is opened.

PR #1300, opened 2024-03-19
Create async-labels when branch or tag is created

benedikt-voelkel:


benedikt-voelkel:
@ktf would that be a problem for you?
To make life easier for maintainers and people responsible for the async. SW requests, we would like to add labels for requesting and accepting automatically, when new branches or tags are pushed.
I tested it in my private repo and it works fine (see https://github.com/benedikt-voelkel/ali-bot/actions/runs/8340720748/job/22825229658)

benedikt-voelkel:
@ktf 

Now focusing on what I called `async-update.yml` for now:
1. It goes through the commits on a `async-<branch-suffix>`
1. if the commit is seen to have `cherry picked from commit` in the commit message, we take that and try to find the original PR
1. basically report (the how to report is TBD)

This is meant to run when a branch is pushed.

Few remaining things to be clarified

Need to find a reliable `break` when scanning through the commits. At the moment this is done whenever the first commit without a linked commit is found. Problem: When a cherry pick has conflicts, the `cherry-pick -x` does **not** add the previous SHA and this condition might not be sufficient. Unless in those cases we do `cherry-pick -e` or in fact amend the commit message once cherry-picked.

Most obvious: Where/how to publish.

Probably things I am forgetting right now.

.github/workflows/async-create-label.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Create async branch or tag label
+        env:
+          GH_TOKEN: ${{ github.token }}
+          ref: ${{ github.ref }}
+
+        id: create_label
+        run: |
+          # to be on the safe side, check the full ref
+          what_am_i=""
+          if [[ "${ref}" != "refs/heads/*" ]] ; then
```
ktf, line 26: are you sure about the `!=`?

ktf: I would actually use a case statement:
```
case "${ref}" in
  refs/heads/*) WHAT_AM_I=branch ;;
  refs/tags/*) WHAT_AM_I=tag ;;
  *) echo "Invalid ref ${ref}, abort." && exit 1 ;;
esac
```

.github/workflows/async-update.yml
```diff
@@ -0,0 +1,82 @@
...
+          GH_TOKEN: ${{ github.token }}
+          ref: ${{ github.ref }}
+
+        id: create_label
+        run: |
+          branch_name="${ref##*/}"
+
+          # we check now if this branch contains the tag <branch-name>-start which is the one where we branched off the default back back in the days
+          start_tag="${branch_name}-start"
+          if ! GIT_DIR=${GITHUB_REPOSITORY} git branch --contains "${start_tag}" | grep "${branch_name}" ; then
```
ktf, line 28: Shouldn't you make sure that branch_name is exactly matched?

.github/workflows/async-update.yml
```diff
@@ -0,0 +1,82 @@
...
+            if [[ "${line}" == "commit *" ]] ; then
+
+              # break if we reached the tag where this branch started
+              current_commit_sha_tmp=$(echo "${line}" | cut -f 2 -d ' ')
+              [[ "${current_commit_sha_tmp}" == "${commit_of_start_tag}" ]] && break
+
+              if [[ -n  "${current_commit_sha}" && -z "${linked_commit_sha}" ]] ; then
+                # this means we ran into a new commit without finding a linked SHA in between, break here
+                echo "Publish only this commit for which we don't find a linked commit"
+                # do something
```
ktf, line 57: are you missing something here?

.github/workflows/async-create-label.yml
benedikt-voelkel: Agreed, and the `!=` was indeed wrong.

.github/workflows/async-update.yml
benedikt-voelkel: Yes. done now

.github/workflows/async-update.yml
benedikt-voelkel: Yes, but the question how/where to publish would be something that we would need to discuss.
This is why I only put comments.

benedikt-voelkel:
We need to discuss again the usage of an `accepted` label, sorry 😬 Let's do that with @chiarazampolli @noferini in person.

The workflows here are supposed to run
```yaml
on:
  push:
    branches:
      - "async-v*"
```
In that way, these WFs are only triggered in very specific cases. This condition needs to be implemented in the downstream repos while of course in `ali-bot` itself it has to be
```yaml
on:
  workflow_call:
````

The last important open point would be to find a solution on where to publish a summary of things that have been included during cherry-picking.
Could be something like a MD table on our O2 GitHub page.

benedikt-voelkel:
Not needed anymore, closing.

PR #1303, opened 2024-04-22
Add possibility to add/remove all labels

benedikt-voelkel:
* also don't mention *-accepted labels

* if "all" or "!all" mentioned, take precedence over specific labels

* "all" takes precedence over "!all"

* do not add and remove the same labels at the same time


benedikt-voelkel:
@ktf to be able to use `+async-label all`

PR #1306, opened 2024-04-25
Publish async tags on slc9

ktf:


ktf:
@chiarazampolli @noferini this should be what is missing to publish the tags.

PR #1333, opened 2024-05-31
Show PR URL in the message of the formatting PR

vkucera:
Users still often miss that a formatting PR has been made to fix their PR. Linking user's PR in the message of the formatting PR automatically makes the reference appear in the conversation of the original PR, which makes the formatting PR more visible to the user.


ktf:
Thank you!

PR #1334, opened 2024-05-31
O2-5010: Use mariadb rather than mysql

ktf:
O2-5010: Use mariadb rather than mysql


ktf:
@adriansev FYI, it should be there in a bit. readline, next week.

adriansev:
Sure! Thanks a lot @ktf !!

PR #1359, opened 2024-06-26
Add alibuild and alidist versions to report-pr-errors log

singiamtel:
Should make it easier to debug issues in the pipeline.

It fails gracefully if either command fails, doesn't interrupt the report

### DIFF
--- a/alibot_helpers/github_utilities.py
+++ b/alibot_helpers/github_utilities.py
@@ -407,8 +407,3 @@ def setGithubStatus(cgh, args, debug_print=True):
         repo_name=repo_name,
         ref=commit_ref
     )
-
-
-def generate_gh_htmllink_from_commit(commit: str, repo: str) -> str:
-    return f"<a href=\"https://github.com/{repo}/commit/{commit}\"\
-    target=\"_blank\">{repo}@{commit}</a>"
--- a/alibot_helpers/github_utilities.py
+++ b/alibot_helpers/github_utilities.py
@@ -407,3 +407,8 @@ def setGithubStatus(cgh, args, debug_print=True):
         repo_name=repo_name,
         ref=commit_ref
     )
+
+
+def generate_gh_htmllink_from_commit(commit: str, repo: str) -> str:
+    return f"<a href=\"https://github.com/{repo}/commit/{commit}\"\
+    target=\"_blank\">{repo}@{commit}</a>"
--- a/alibot_helpers/utilities.py
+++ b/alibot_helpers/utilities.py
@@ -2,8 +2,6 @@
 
 import shlex
 import sys
-from os import devnull
-import subprocess
 
 
 def parse_env_file(env_file_path):
@@ -25,11 +23,3 @@ def to_unicode(s):
     elif not isinstance(s, unicode):
         return unicode(str(s))
     return s
-
-
-def getout(cmd):
-    with open(devnull) as dn:
-        p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=dn, shell=True)
-        out = p.communicate()[0].decode('utf-8')
-        code = p.returncode
-    return (out, code)
--- a/alibot_helpers/utilities.py
+++ b/alibot_helpers/utilities.py
@@ -2,6 +2,8 @@
 
 import shlex
 import sys
+from os import devnull
+import subprocess
 
 
 def parse_env_file(env_file_path):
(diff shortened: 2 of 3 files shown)


singiamtel:
Example output 

image

singiamtel:
New format (added a link)

image

report-pr-errors
```diff
@@ -21,7 +21,7 @@ from alibot_helpers.github_utilities import (
     calculateMessageHash, setGithubStatus, parseGithubRef, GithubCachedClient,
     PickledCache,
 )
-from alibot_helpers.utilities import to_unicode
+from alibot_helpers.utilities import to_unicode, getout
```
ktf, line 24: maybe a better approach would be to expose the function in alibuild to retrieve the version?

report-pr-errors
```diff
@@ -314,6 +317,27 @@ class Logs(object):
...
+            (alibuild_version, status) = getout("alibuild version")
+            if status:
+                self.alibuild_version = "Could not get alibuild version"
+            else:
+                self.alibuild_version = alibuild_version
+        except Exception as e:
+            print("Error while getting alibuild version:", e, file=sys.stderr)
+            self.alibuild_version = "Could not get alibuild version"
+        try:
+            (alidist_version, status) = getout("git -C ./alidist rev-parse --short HEAD")
```
ktf, line 331: Do not assume alidist is using git.

ktf (changes requested):
Thank you, I was thinking about something similar actually. I am not sure of the  approach though. Could you check the comments?

alibot_helpers/utilities.py
```diff
@@ -23,3 +25,11 @@ def to_unicode(s):
     elif not isinstance(s, unicode):
         return unicode(str(s))
     return s
+
+
+def getout(cmd):
```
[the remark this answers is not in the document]
singiamtel: Using subprocess.run instead

report-pr-errors
singiamtel: I'm not sure I get what you mean. Currently in ali-bot we only pickup whatever alibuild we find in $PATH, you'd like to import it as a python module instead?

report-pr-errors
singiamtel: It now fails when alidist is not a git repo. I'm not sure how to get the alidist version when it doesn't have history though

```bash
$ git --git-dir ./alidist/.git rev-parse --short HEAD
f0f86990
$ mv ./alidist/.git ./alidist/.git2
$ git --git-dir ./alidist/.git rev-parse --short HEAD
fatal: not a git repository: './alidist/.git'
```

﻿Screenshot 2024-06-26 at 14 33 47 image

ktf: Why not simply passing the alibuild version (obtained via `aliBuild version`) and the alidist one as parameters / environment variables? This way you do not need to know how to get them from the script reporting about the builds. You could actually reuse ALIBUILD_SLUG and ALIDIST_SLUG which I think we already define.

singiamtel: Are they also set for CI jobs? I've only found the definition for the Jenkins jobs.

(if they aren't we can also add them before merging here and just use those)

singiamtel: Implemented getting the version from the env vars, as discussed

alibot_helpers/github_utilities.py
```diff
@@ -407,3 +407,8 @@ def setGithubStatus(cgh, args, debug_print=True):
         repo_name=repo_name,
         ref=commit_ref
     )
+
+
+def generate_gh_htmllink_from_commit(commit: str, repo: str) -> str:
```
ktf, line 412: If it's not used elsewhere, I would prefer we define it only where it's needed.

report-pr-errors
```diff
@@ -314,6 +317,22 @@ class Logs(object):
...
+        # The versions are defined differently for both job runner setups
+        # Jenkins (daily builds):
+        #   - aliBuild: ALIBUILD_SLUG
+        #   - alidist: ALIDIST_SLUG (if not set it defaults to alidist@master)
+        # Nomad (CI checks):
+        #   - aliBuild: INSTALL_ALIBUILD
+        #   - alidist: (defaults to alidist@master)
+        self.alibuild_version = os.getenv('ALIBUILD_SLUG') or os.getenv('INSTALL_ALIBUILD') or 'Not pinned (most likely alibuild@master)'
+        alidist_version = os.getenv('ALIDIST_SLUG')
+        if alidist_version:
```
ktf, line 330: I am not sure I understand the logic here. Why do you do two different things depending on if it is alibuild_version or alidist version?

singiamtel: This is an old change from when I was getting the commit hash, which was easy to convert to a github URL. The new version is not, so I'll remove the github links

singiamtel:
Should be fixed now

ktf (approved):

PR #1361, opened 2024-07-01
Move to a clang-format which knows about C++20

ktf:


ktf:
Move to a clang-format which knows about C++20.

PR #1362, opened 2024-07-03
Bump alibuild to 1.17.7 on centos&ubuntu repos

singiamtel:
This version has been available for a while and seems stable. The most
notable change is managing the git repos with --filter=tree:0, which
should mean a notably smaller sw/ folder size

Setting it to draft as centos7 is now EOL, so the action might fail when trying to deploy to those repos

### DIFF
--- a/.github/workflows/o2-full-deps.yml
+++ b/.github/workflows/o2-full-deps.yml
@@ -43,7 +43,7 @@ jobs:
     env:
       ALIBUILD_TAG: ${{ matrix.alibuild_tag }}
       DISTRO: ${{ matrix.el_version }}
-      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: 'true'
+      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: 'true' # Only required for CentOS 7
 
     steps:
       # For rpms/*.spec
--- a/.github/workflows/o2-full-deps.yml
+++ b/.github/workflows/o2-full-deps.yml
@@ -35,7 +35,7 @@ jobs:
             container: almalinux:9
             alibuild_tag: v1.17.7
           - el_version: fedora
-            container: fedora:33
+            container: fedora:40
             alibuild_tag: v1.17.7
 
     name: RPM (${{ matrix.el_version }})
--- a/.github/workflows/o2-full-deps.yml
+++ b/.github/workflows/o2-full-deps.yml
@@ -43,11 +43,11 @@ jobs:
     env:
       ALIBUILD_TAG: ${{ matrix.alibuild_tag }}
       DISTRO: ${{ matrix.el_version }}
+      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: 'true'
 
     steps:
       # For rpms/*.spec
       - uses: actions/checkout@v3
-
       - name: Install prerequisites
         run: |
           set -ex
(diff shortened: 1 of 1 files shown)


adriansev:
@singiamtel is alma9 included in this? as in the el9 repository the latest alibuild version is 1.16.1

adriansev:
oh, my bad, i saw in changes that is it... but also note that fedora33 is not among us anymore, it would be best if fedora is to be supported to use fedora 40

singiamtel:
> oh, my bad, i saw in changes that it is... but also note that fedora33 is not among us anymore, it would be best if fedora is to be supported to use fedora 40

Sure, using fedora 40 now

> @singiamtel is alma9 included in this? as in the el9 repository the latest alibuild version is 1.16.1

Yeah, I'd prefer to keep the same alibuild version for all repos if possible

.github/workflows/o2-full-deps.yml
```diff
@@ -64,14 +64,14 @@ jobs:
           set -x
           rpmbuild -ba "rpms/o2-prereq-$DISTRO.spec"
           git clone -b "$ALIBUILD_TAG" https://github.com/alisw/alibuild
-          cd alibuild
+          cd alibuild || exit
```
ktf, line 67: Did you check if it is running with `set -e`?

.github/workflows/o2-full-deps.yml
```diff
@@ -64,14 +64,14 @@ jobs:
...
           case "$DISTRO" in
             el7) deps=python3,git,python3-PyYAML,python3-requests,python3-distro ;;
             *)   deps=python3,git,python3-pyyaml,python3-requests,python3-distro ;;
           esac
           python3 setup.py bdist_rpm --requires "$deps"
-          cp dist/alibuild-*.noarch.rpm ~/rpmbuild/RPMS/$(uname -m)/
-          cd  ~/rpmbuild/RPMS/$(uname -m)/
+          cp dist/alibuild-*.noarch.rpm ~/rpmbuild/RPMS/"$(uname -m)"/
+          cd  ~/rpmbuild/RPMS/"$(uname -m)"/ || exit
```
ktf, line 74: Same as above.

.github/workflows/o2-full-deps.yml
singiamtel: That is a cleaner fix indeed, and actionlint doesn't seem to complain. It's now implemented

singiamtel:
Just tested the centos7 action, and it fails with the same yum repo missing issue. Should we delete it?

ktf:
Can we fix it using vault.cern.ch? I am afraid we still have a few things on Centos7.

singiamtel:
> Can we fix it using vault.cern.ch? I am afraid we still have a few things on Centos7.

Sure, did the workaround

I also had an issue with the Node runner (https://github.com/actions/runner/issues/2906), there is a workaround but we should get it of it as soon as we can drop CentOS 7 support

ktf:
We also need the workaround for pip, apparently. See https://github.com/alisw/ali-bot/actions/runs/9805778473. Can you take care of it?

singiamtel:
> We also need the workaround for pip, apparently. See [alisw/ali-bot/actions/runs/9805778473](https://github.com/alisw/ali-bot/actions/runs/9805778473). Can you take care of it?

@kft Haven't managed to fix this yet, the python version on their repos is just too old for the new pip. Building python from source would probably work but it means we'd get the compiler toolchain on the image, and the action would take considerably longer to run too

How should we proceed?

PR #1379, opened 2024-08-06
Export variables for current version running in CI

singiamtel:
Properly pass down aliBuild/alidist/ali-bot env variables for logging

Should fix #1359 when running in CI


ci/continuous-builder.sh
```diff
@@ -87,6 +87,9 @@ if [ -n "$HASHES" ]; then
       # Setup environment. Skip build if the .env file doesn't exist any more.
       source_env_files "$env_name" || exit
 
+	  # Pass the versions down for logging
```
ktf, line 90: Is the indentation right?

singiamtel: I was using tabs by mistake, should be fixed now

PR #1384, opened 2024-08-09
Create daily tags inside venv

singiamtel:
Needed for Ubuntu 24.04, otherwise fails because of the new "externally managed envs" (https://peps.python.org/pep-0668/)

I can look into how to make it specific for Ubuntu 24.04 only if you'd rather not have this on all daily releases

Tagging @ktf

### DIFF
--- a/jenkins/daily-tags
+++ b/jenkins/daily-tags
@@ -32,6 +32,8 @@ try {
               rm -rf "$PYTHONUSERBASE"
               yum install -y python3-pip python3-devel python3-setuptools
               mkdir -p /local/tmp
+              python -m venv /local/tmp/venv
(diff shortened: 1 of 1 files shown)


jenkins/daily-tags
```diff
@@ -32,6 +32,8 @@ try {
               rm -rf "$PYTHONUSERBASE"
               yum install -y python3-pip python3-devel python3-setuptools
               mkdir -p /local/tmp
+              python -m venv /local/tmp/venv
```
ktf, line 35: Since this is only for ubuntu, I would protect it with a `case $ARCHITECTURE in ubuntu24*)...`.

PR #1401, opened 2024-09-19
Delete SLC7 alidist checker

singiamtel:
Superseded by ~build/O2/alidist-slc9-aarch64~ build/O2/alidist-slc9-x86


ktf:
I guess you mean x86-64?

PR #1412, opened 2024-10-07
Use defaults o2 for AliRoot CI

ktf:


ktf:
To be merged after the daily build is available.

PR #1416, opened 2024-10-16
Make sure AliRoot builds with o2 defaults

ktf:


ktf:
@davidrohr @pzhristov as discussed in the meeting.

PR #1421, opened 2024-11-13
DPG: Bump to SLC9

singiamtel:
In principle there's no reason to keep them in SLC7 seeing that the production workload happens in SLC9. 

Bumping should also help narrow down the builder error on https://github.com/AliceO2Group/O2DPG/pull/1789 

@ktf would it be okay to also migrate `build/O2Physics/o2` to SLC9? In that case we could convert the last `slc7-o2physics` CI builder to SLC9 too

CC @chiarazampolli @jackal1-66


ktf:
> @ktf would it be okay to also migrate build/O2Physics/o2 to SLC9? In that case we could convert the last slc7-o2physics CI builder to SLC9 too

I thought we already did a while ago...

PR #1428, opened 2024-11-29
Create parallel FullCI testers in slc9

singiamtel:
CC @ktf @davidrohr


ktf:
Looks good. Maybe we should remove the CC8 ones as well?

PR #1445, opened 2024-12-11
Cleanup alienv file

singiamtel:
Patch mostly generated with `shellcheck -f diff cvmfs/alienv | git apply` and 
then manually removed some unused variables

I want to do several improvements here, but seeing as it's a sensitive script,
I'd like to do them gradually

I haven't been able to find behaviour differences between this and the one
in CVMFS

CC @ktf 

I will reflect the change in CVMFS after merge

### DIFF
--- a/cvmfs/alienv
+++ b/cvmfs/alienv
@@ -3,54 +3,53 @@
 version=3.2.10
 cvmfsdir=/cvmfs/alice.cern.ch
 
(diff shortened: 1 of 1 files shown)


ktf (changes requested):
Too much of a change. Please split it in multiple PRs.

PR #1454, opened 2025-01-17
Don't require spaces in makefiles

singiamtel:
As reported by @davidrohr 

Should fix https://github.com/AliceO2Group/AliceO2/actions/runs/12818036733/job/35742588079?pr=13865

### DIFF
--- a/.github/workflows/c++-code-formatting.yml
+++ b/.github/workflows/c++-code-formatting.yml
@@ -342,8 +342,11 @@ jobs:
                     "Remove the trailing spaces at the end of the line.")
             }
             match($0, /\t+/) {
-              error("Tab characters found",
-                    "Indent code using spaces instead of tabs.")
+              # Only check for tabs in C/C++ source files
+              if (FILENAME ~ /\.[ch](xx|pp)?$/) {
+                error("Tab characters found",
+                      "Indent code using spaces instead of tabs.")
+              }
             }
             END {
               exit exit_code


.github/workflows/c++-code-formatting.yml
```diff
@@ -342,8 +342,10 @@ jobs:
                     "Remove the trailing spaces at the end of the line.")
             }
             match($0, /\t+/) {
-              error("Tab characters found",
-                    "Indent code using spaces instead of tabs.")
+              if (FILENAME !~ /[Mm]akefile/) {
```
ktf, line 345: I would do it the other way around. I would only complain about tabs if they are found in .c / .cxx files. There is plenty of cases where having tabs might actually be justified.

.github/workflows/c++-code-formatting.yml
```diff
@@ -342,8 +342,11 @@ jobs:
                     "Remove the trailing spaces at the end of the line.")
             }
             match($0, /\t+/) {
-              error("Tab characters found",
-                    "Indent code using spaces instead of tabs.")
+              # Only check for tabs in C/C++ source files
+              if (FILENAME ~ /\.[ch](xx|pp)?$/) {
```
ktf, line 346: Probably `.py` and `.sh`  as well.

ktf:
Merging. Please take into account the comment, though.

PR #1459, opened 2025-01-23
Sync alienv to CVMFS automatically

singiamtel:
I tried to make it as crash resistant as possible, so hopefully it doesn't
interfere with the regular publishing

CC @ktf

### DIFF
--- a/publish/aliPublishS3
+++ b/publish/aliPublishS3
@@ -16,6 +16,7 @@ from tempfile import NamedTemporaryFile, mkdtemp
 from subprocess import Popen, PIPE, STDOUT, DEVNULL, getstatusoutput
 from smtplib import SMTP
 from xml.etree.ElementTree import ElementTree
+from pathlib import Path
 
 def format(s, **kwds):
   return s % kwds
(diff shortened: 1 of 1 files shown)


ktf:
As discussed, make sure you back up the old version before deployment of the new one, so that people can use the "safe" version in case of accident.

publish/aliPublishS3
```diff
@@ -769,6 +771,24 @@ def sync(pub, architectures, s3Client, bucket, baseUrl, basePrefix, rules,
...
 
+  # Update the alienv script in CVMFS if it has changed
+  try:
+    new_alienv = dirname(realpath(__file__)) / ".." / "cvmfs" / "alienv"
+    old_alienv = "/cvmfs/alice.cern.ch/bin/alienv"
+    
+    # Create backup with timestamp if files are different
+    if not filecmp.cmp(new_alienv, old_alienv, shallow=False):
+        timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+        backup_alienv = f"/cvmfs/alice.cern.ch/bin/alienv.safe-{timestamp}"
```
ktf, line 782: Maybe we should have a separate folder for the backups (or limit the backups to just a few).

ktf (changes requested):
Please followup the remaining comments.

PR #1475, opened 2025-02-10
Tweak some scripts to get them running under macOS

singiamtel:
`ciqueues` is broken when running a pip installed version, but it works when using a local clone

### DIFF
--- a/utils/ci-status-history
+++ b/utils/ci-status-history
@@ -15,9 +15,20 @@ fi
 
 [ -z "$check" ] && show_check=true
 
+# The temporary file lets us format the output nicely
+tmpfile=$(mktemp)
+trap 'rm -f $tmpfile' EXIT
+
+if [ "$show_check" = true ]; then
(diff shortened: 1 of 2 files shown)


utils/ci-status-history
```diff
@@ -15,9 +15,20 @@ fi
...
+else
+  printf "DATE\tTIME\tUSER\tSTATUS\tDESCRIPTION\n" > "$tmpfile"
+fi
+
 curl -fsS "$api_url" | jq -r "[.[]
    ${check:+| select(.context == \"$check\")}
    | [(.created_at | sub(\"T\"; \"\\t\")), .creator.login, .state, ${show_check+.context,} .description]
    | join(\"\\t\")
-] | sort | .[]" |
-  column -tN "DATE,TIME,USER,STATUS,${show_check+CHECK,}DESCRIPTION" -s '	'  # literal tab char!
```
ktf, line 23: This should work on macOS as well... Maybe `${show_check:+CHECK,}` though?

ktf: Maybe we would be better off using @tsv in jq.

PR #1476, opened 2025-02-12
Publish el9 RPMs too

singiamtel:
As requested by @ehellbar

CC @ktf


ktf:
I think the updatable RPMS are not needed anymore. @Barthelemy @vascobarroso can you confirm?

PR #1481, opened 2025-02-14
Improve error message

singiamtel:
This happens often in AliPhysics

We should probably reduce the number of updates to the commits somehow
at some point


ktf (approved):

ktf:
Does this mean that we set the status even if it's the same? Indeed that should not happen since it's eating our quota for nothing.

PR #1482, opened 2025-02-14
Replace old aliswmod references with alienv

singiamtel:


ktf:
Please follow up with @Barthelemy about this one. Is aliswmod still used?

Barthelemy:
those are just comments and email notifications right ? 
It seems fine to me

ktf:
I think aliswmod was at some point used in some FLP instructions. Is it still the case?

vascobarroso:
We had it in the past but I think it's all gone. 
On `ControlWorkflows` it was removed [here](https://github.com/AliceO2Group/ControlWorkflows/pull/26).

@knopers8, could you just confirm that AliECS doesn't use it in any other place ?

ktf:
@knopers8 can you please comment here?

knopers8:
Sorry, I missed your comments before. There are no uses of `aliswmod` in AliECS, you can go ahead.

PR #1485, opened 2025-02-17
Update link_latest_gcc script

singiamtel:
Default to new GCC, and improve some sharp edges

Now it warns about several old architectures where we don't publish anymore. Are they still used anywhere, or would it be okay to delete them altogether?

```
Warning: No GCC v13.2.0 toolchain found for arch el5-x86_64
Warning: No GCC v13.2.0 toolchain found for arch el6-x86_64
Warning: No GCC v13.2.0 toolchain found for arch el7-aarch64
Warning: No GCC v13.2.0 toolchain found for arch ubuntu1404-x86_64
Warning: No GCC v13.2.0 toolchain found for arch ubuntu1604-x86_64
```

### DIFF
--- a/cvmfs/link_latest_gcc.sh
+++ b/cvmfs/link_latest_gcc.sh
@@ -1,37 +1,25 @@
-#!/bin/bash
-set -eo pipefail
+#!/bin/bash -e
+set -o pipefail
 GCC_VER=${1:-"v13.2.0"}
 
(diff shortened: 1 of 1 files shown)


ktf:
I think this would be better off if we moved it to the publisher, so that the Toolchain is there in the same transaction when a new architecture is created.

PR #1507, opened 2025-03-17
Remove old script

singiamtel:


singiamtel:
> Did you check it's not in the setup.py?

Yeah I did, and in principle we never have to worry about that again as it would make the `pip install` tests fail

PR #1515, opened 2025-03-31
Fix shellcheck warning

singiamtel:
This is a small change picked from #1445 to test whether #1459 is
properly creating the backups of the original alienv

CC @ktf


ktf:
I suggest we merge this kind of things in the morning. Apart from that, looks fine.

PR #1517, opened 2025-04-07
Add `bc` to slc9 dependencies

singiamtel:


ktf:
Can't we get rid of bc usage?

singiamtel:
We would need to chase all places where it's needed and replace it, I wanted to add it because of these ones, but there are probably more

https://github.com/AliceO2Group/AliceO2/blob/cadc5fa33d7eb8f28551a5f83c65d833d97ced16/Utilities/Tools/jobutils.sh#L294
https://github.com/AliceO2Group/AliceO2/blob/cadc5fa33d7eb8f28551a5f83c65d833d97ced16/prodtests/full-system-test/calib-workflow.sh#L18

PR #1520, opened 2025-04-14
Fix linter issues: Imports

singiamtel:
Unused imports, cleanup comparisons, etc


ktf:
As usual, I would prefer we split this kind of changes into one PR per kind of fix.

PR #1524, opened 2025-04-14
Fix linter issues: Unused f strings

singiamtel:
Includes prefer def over lambda, unused variables, unused f strings

### DIFF
--- a/alibot_helpers/github_utilities.py
+++ b/alibot_helpers/github_utilities.py
@@ -252,7 +252,7 @@ class GithubCachedClient(object):
         response = requests.post(url=url, data=data, headers=headers)
         sc = response.status_code
         if sc == 422:
-            print(f"GitHub error: Unprocessable Entity", file=sys.stderr)
+            print("GitHub error: Unprocessable Entity", file=sys.stderr)
             print('This usually means that the commit has reached the maximum number of statuses.', file=sys.stderr)
         return sc
 


cvmfs-mon/cvmfs-mon.py
```diff
@@ -42,7 +42,7 @@ def check(monit):
     try:
       api_url = monit["api_url"] + "/" + repo
       s = get(api_url).json()
-      ok = s["status"] == "ok"
```
ktf, line 45: What is the point of the comparison in the first place? Please split also this one.

ci/process-pull-request-http.py
```diff
@@ -172,7 +172,7 @@ def action_check_permissions(self, git, pr, perms, tests):
     self.action_approval_required(git, pr, perms, tests)
 
   def action_approval_required(self, git, pr, perms, tests):
-    pull = git.get_pull(pr, cached=True)
```
ktf, line 175: also here, I am not sure why you need to get the PR if then it's not needed. Does it update the cache?

alibot_helpers/github_utilities.py
```diff
@@ -252,7 +252,7 @@ def post(self, url, data, stable_api=True, **kwds):
         response = requests.post(url=url, data=data, headers=headers)
         sc = response.status_code
         if sc == 422:
-            print(f"GitHub error: Unprocessable Entity", file=sys.stderr)
```
ktf, line 255: This is fine. Just split it in a separate PR.

PR #1525, opened 2025-04-14
Remove unused script

singiamtel:


cvmfs-mon/cvmfs-mon.py
```diff
@@ -44,7 +44,6 @@ def check(monit):
     try:
       api_url = monit["api_url"] + "/" + repo
       s = get(api_url).json()
-      ok = s["status"] == "ok"
```
ktf, line 47: On a second thought. What happens if s does not have "status"? Is that actually possible at all?

singiamtel: As far as I can tell, this script is not running anywhere so I can't verify what the API returns. Maybe it'd be better to just delete it?

ktf: > Maybe it'd be better to just delete it?

Yes...

PR #1526, opened 2025-04-14
Bump aliBuild to v1.17.17

singiamtel:


ktf:
I would start with one arch as usual, and then do it everywhere.

davidrohr:
Then let's start with slic9-gpu builder for the FUllCI :)

PR #1530, opened 2025-05-07
Remove unused variable

singiamtel:


ci/process-pull-request-http.py
```diff
@@ -174,7 +174,6 @@ def action_check_permissions(self, git, pr, perms, tests):
     self.action_approval_required(git, pr, perms, tests)
 
   def action_approval_required(self, git, pr, perms, tests):
-    pull = git.get_pull(pr, cached=True)
```
ktf, line 177: Maybe the git pull has a side effect?

singiamtel: As far as I can tell, no important ones. It stores it in a cache, but it would be basically the same to just make the request when we actually need it

PR #1538, opened 2025-05-21
Fix typo on aliPublish

singiamtel:
CC @sawenzel


singiamtel:
Currently broken, merging

PR #1541, opened 2025-05-30
Standardize on pyproject.toml

singiamtel:
Follow up to #1506, this should mirror the current setup.py behavior in
relationship to bundling the scripts in the package


ktf:
@singiamtel why this was not merged? Are there issues with it?

singiamtel:
No issues, I think we just forgot about it. It's okay for merging imo

PR #1545, opened 2025-06-18
o2-prereq-el9: Add subversion

singiamtel:
Apparently required for the dataflow build


ktf:
Seriously? Why?

singiamtel:
@ktf could we merge this? The dataflow builds are broken otherwise

PR #1547, opened 2025-06-23
Rename cs8 jobs

singiamtel:
Oversight from #1531

Needs required checks to be renamed right after merge, or the CI will be stuck


ktf:
We should probably have some expansion for the folder name where the checks are located. Something like $CHECK_ARCH.

PR #1548, opened 2025-06-23
Do full CI for ARM

ktf:
So that we can finally test if it works.


ktf:
@singiamtel @davidrohr any objections?

singiamtel:
No objections from my side, although I expect we'll find some currently broken packages on ARM on the way

davidrohr:
Did you test this before? FullCI will e.g. run the GPU RTC compilation test. I assume this will not work for ARM, since it requires CUDA and HIP compilers in the build container.

PR #1552, opened 2025-07-14
Make sure special characters are escaped when using requests

ktf:
This should fix the mismatch between # being correctly handled by boto3,
but not by a direct request via requests.


ktf:
Verified to work correctly for gpu-system. Merging.

PR #1556, opened 2025-07-14
Fix unbound variable

singiamtel:
There's some case where this variable is not set properly, and it
crashes the script

### DIFF
--- a/update-symlink-manifests
+++ b/update-symlink-manifests
@@ -168,6 +168,7 @@ def build_package_manifest(s3c, bucket, package, read_only):
     """Create a symlink manifest for a single package."""
     symlinks = {}
     manifest = package.rstrip("/") + PACKAGE_MANIFEST_EXT
+    link_key = None
 
     # First, fetch the existing manifest (if any) for this package.
     try:


update-symlink-manifests
```diff
@@ -168,6 +168,7 @@ def build_package_manifest(s3c, bucket, package, read_only):
     """Create a symlink manifest for a single package."""
     symlinks = {}
     manifest = package.rstrip("/") + PACKAGE_MANIFEST_EXT
+    link_key = ""
```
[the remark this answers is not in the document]
singiamtel: Sounds good

update-symlink-manifests
```diff
@@ -168,6 +168,7 @@ def build_package_manifest(s3c, bucket, package, read_only):
     """Create a symlink manifest for a single package."""
     symlinks = {}
     manifest = package.rstrip("/") + PACKAGE_MANIFEST_EXT
+    link_key = None
```
ktf, line 171: This is probably fine. Why would link_key be unbound, though?

ktf: actually, I do not see how it can be unbound.

singiamtel: It happens on this branch


```python
    except botocore.exceptions.ClientError as err:
        # Treat a missing manifest like an empty one; i.e., use only the
        # individual symlinks.
        LOG.info("error while fetching %s: %s; recreating from scratch",
                 manifest, err)
     else: # only runs when there's no exception
        LOG.info("found existing manifest %s", manifest)
        for i, line in enumerate(lines):
            link_key, sep, target = line.partition("\t") 
```

ktf:
I would claim #1557 is the correct fix.

PR #1573, opened 2025-07-28
Remove old trusted users

singiamtel:


ktf:
You should probably add yourself.

Barthelemy:
Yes, it needs to be updated. 

I propose : 
qc: add Piotr
config and monitoring: add Matteo
Readoutcard: add Sylvain

ktf:
For the record, Piotr => @knopers8 , Sylvain => @sy-c, Matteo @mconcas

PR #1579, opened 2025-08-22
Remove unused vars

singiamtel:
As far as I can tell, the only one that's used is ALIBOT_PR_REPO, in the
alidist o2.sh recipe


ktf:
Are you sure? That seems odd.

PR #1583, opened 2025-09-01
Update license metadata

singiamtel:
Follow up to https://github.com/alisw/alibuild/pull/961

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,12 +11,12 @@ description = 'ALICE Multipurpose bot'
 authors = [
   {name = 'Giulio Eulisse', email = 'giulio.eulisse@cern.ch'},
 ]
-license = {text = 'GPL'}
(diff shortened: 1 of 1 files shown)


singiamtel:
Yes, exactly. It's the same, I get it on aliBuild too when trying to run `uv run --python 3.8 pip install .`

singiamtel:
> Ok, but why? Shall we just drop python 3.8 support? What does slc7 use?

Python 3.6 apparently. We didn't update further because the slc7 repositories didn't have any higher version, we'd need to build from source. 

Should we install python via `uv` in the containers? They provide binaries for 3.8 and above, we could standardize all containers on the same version

ktf:
I would be happy simply by getting 3.8 working, then.

singiamtel:
Closing following https://github.com/alisw/alibuild/pull/965

PR #1657, opened 2026-03-11
Run smoke test with uv

singiamtel:
I would like to eventually switch our CI and daily builds to uv, as it gives us much more control on the version and environment (and allows us to remove several hacks like PYTHONPATH)

Starting with the smoke test to make sure it works


ktf:
uv got bought by OpenAI. I would say we stick with pip.

PR #1681, opened 2026-04-01
Add prerequisites for el10

ktf:


ktf:
@davidrohr @vascobarroso @Barthelemy next time we run this should also produce the prerequisite RPMS for Alma10

PR #1686, opened 2026-05-04
Fix architecture of staging O2Physics CI checker

vkucera:
The container architecture has to be in the `arch` field of the Nomad job configuration which then (together with the suffix `-staging`) has to match the path of the `.env` checker configuration file in the `ali-bot` repo.


ktf:
Please write here what you told me privately about the issue.

vkucera:
> Please write here what you told me privately about the issue.

Description updated

PR #1687, opened 2026-05-08
Pass RECC_CONFIG_DIRECTORY to aliBuild

vkucera:
If using RECC inside a container and with defined `RECC_CONFIG_DIRECTORY`, mount the directory path inside the container at an internal `RECC_CONFIG_DIRECTORY` path, otherwise pass it as is.


ktf:
Merging, but avoid the boilerplate in a next PR.

PR #1692, opened 2026-05-26
Use RECC in build/O2Physics/o2

vkucera:


ktf:
Any reason for not having `RECC_LOG_LEVEL=error`? Fine with me otherwise.

vkucera:
> Any reason for not having `RECC_LOG_LEVEL=error`? Fine with me otherwise.

Just consistency with `build/alidist/O2Suite/o2/macOS-arm/staging`. What do you prefer?

ktf:
if it works, we should move everything to error (or warn).

PR #1700, opened 2026-07-20
Allow checking out PR head in code formatting

vkucera:
Starting from today, all versions of checkout forbid checking out fork pull request code on `pull_request_target` by default.


ktf:
We should properly reevaluate this.

PR #1721, opened 2026-08-17
Add tests for list-branch-pr

ktf:
Characterisation tests pinning what the CI builders depend on: the five-field
output contract, untested-before-rebuilds selection, that the hash shards
partition the pull requests, and that --no-status keeps the listing read-only.
They assert today's behaviour rather than an ideal, so changing it has to be
deliberate.

Also runs them: test/ had no workflow, so the existing test was never executed.


ktf:
Test harness.

PR #1731, opened 2026-08-20
Use RECC by default if available

vkucera:
- Add RECC variables in `DEFAULTS.env` for all checkers that support it.
- Remove redundant RECC variables from configurations of tests in those checkers.


ci/repo-config/mesosci/slc9-staging/DEFAULTS.env
```diff
@@ -25,3 +25,5 @@ CONTAINER_IMAGE=registry.cern.ch/alisw/slc9-builder:latest
 # Inherited from mesosci/DEFAULTS.env today, restated here so a change to the
 # shared default does not silently retune this pool.
 JOBS=24
+USE_RECC=1
+RECC_LOG_LEVEL=info
```
ktf, line 29: This gets very verbose actually. Are you sure?

vkucera: That's what we have now on the staging machines for both macOS and Alma 9. I was just missing the removal from the original location. Now updated.

